PR #6, opened 2016-07-15
python slot generator

nemik:
I wanted a way to generate the custom slots but without the JS web-app, figured it would be easy enough from python which is already used. adding this for others if it helps.


jingai:
This needs to be modified to support MUSICARTISTS and MUSICPLAYLISTS.  It's still using the old MUSIC slot.

jingai:
I'll update it.. give me a few min.

jingai:
Done with https://github.com/m0ngr31/kodi-alexa/pull/13

PR #14, opened 2016-10-02
Fix WSGI return.

jingai:
Refactoring for Lambda broke WSGI return.


m0ngr31:
Thanks for doing this.

PR #24, opened 2016-10-10
Add support for Pictures.

jingai:
This PR should only really affect the Pictures library, but it's a big commit.

You might notice the references to a certain movie based on a certain Philip K Dick novel.. have been wanting to do this for ages.

But, since it required so many additions I figured I'd run it by you as a PR first.

### DIFF
--- a/alexa.intents
+++ b/alexa.intents
@@ -188,6 +188,78 @@
       "slots": []
     },
     {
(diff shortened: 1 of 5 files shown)


m0ngr31:
Looks good to me. Let's merge it.

PR #26, opened 2016-10-11
script to move Lambda files into a new folder before deploy 

digiltd:
a simple bash script to prevent all files in the folder being uploaded to aws


jingai:
84a5a9e4aa5fbb2097b5556ce86bc428b1e4fc07 shouldn't be part of this PR.

PR #28, opened 2016-10-11
script to move relevant files to folder and lambda-deploy from there

digiltd:
This prevents all the extra files from being uploaded when you deploy

### DIFF
--- a/package-for-lambda.sh
+++ b/package-for-lambda.sh
@@ -1,0 +1,71 @@
+#!/bin/bash   
+
+# By default lambda-deploy will package the entire folder it is run from and
+# upload it to AWS. To keep things a little neater you can run this script to
+# copy the Lambda relevant files to a seperate folder then deploy from the new 
+# folder.
+# 
+# Usage
+# 
+# $ sh package-for-lambda.sh  
+# 
+# Options
+# 
+# --name FOLDER_NAME	The name of the folder you want to package into, note
+# 						that this will also be the name of the Lambda created 
+# 						on AWS. Defaults to kodi-alexa-lambda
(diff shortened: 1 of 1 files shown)


m0ngr31:
Before I download and play with this, does this delete the created folder? If not, should probably add it to the `.gitignore` file. What's the benefit of using it over just lambda-deploy, save on a few files?

digiltd:
Running the lambda-deploy command as it stands uploads the entire kodi-alexa folder with all the Docker, intents, utterances etc etc, plus the .git folder

There is the option to remove the folder at the start of the script, but I didn't put it in automatically as I thought it would have to install the requirements each time you ran it. But that doesn't appear to be the case.

So I will change it to tidy up after itself. 

You're right that is it isn't doing much, it initially was my way of being able to keep things tidy my end without messing too much with the structure that is used for multiple server methods.

m0ngr31:
I didn't think about the `.git` directory or anything. That alone is probably worth it. Do you know anything about batch scripting? I would hate to leave Windows users in the dark using this. Maybe the best solution would be a Python script so it'd be universal.

digiltd:
Sorry, no idea about Windows. I just had a go at trying to work out how to do it in Python, but decided life is too short.

How about a section in the README about it, or a wiki entry? (I could knock something up)

After all it is not too much effort to ask the user to move the four files (`.env kodi.py wsgi.py requirements.txt`) into a new folder before running the `lambda-deploy deploy` command considering everything else they would have done beforehand :)

m0ngr31:
I'll whip something up in python, just to be inclusive.

m0ngr31:
This is being replaced with a Python version in PR #46

PR #29, opened 2016-10-11
add basic step/seek forward and backward

digiltd:
step forward
step backward
big step forward
big step backward

### DIFF
--- a/alexa.intents
+++ b/alexa.intents
@@ -38,6 +38,22 @@
       "slots": []
     },
     {
+      "intent": "StepForward",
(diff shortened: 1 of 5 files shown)


m0ngr31:
Can you update the utterances.txt file to have the new utterances as well? Using [this tool](http://www.makermusings.com/amazon-echo-utterance-expander/) to create the alexa.utterances file.

digiltd:
Ok, have done. I also added a bit of info about what the uttereance.txt file is for.

PR #30, opened 2016-10-12
Only conditionally include 'card' and 'reprompt' in response.

jingai:
Was producing empty cards with only the prefix "SessionSpeechlet - " in
the title and text for every query.  I think we should only produce cards at all if there is additional information that isn't feasible to provide via outputSpeech.

@m0ngr31 , I'm making this a PR because it may affect the Lambda handler, since you changed it to work this way for that.  If it breaks the Lambda handler, I can work up another solution.  But, it's spamming empty cards as it is for the WSGI handler :)

But what I've done is to only include the card if title _and_ output is specified, and I've removed the "SessionSpeechlet - " prefix from both.  I don't know if this was something the Lambda handler made use of, though.

Additionally, outputSpeech is only filled out if there is something for Alexa to say.  Same for reprompt.

There's also a tiny bit of refactoring (utilizing on_intent() in the WSGI handler) and reorganization to group like functions.

### DIFF
--- a/wsgi.py
+++ b/wsgi.py
@@ -51,7 +51,44 @@ import kodi
 INI_FILE = os.path.join(os.path.dirname(__file__), "wsgi.ini")
 
 
-# This utility function constructs the required JSON for a full Alexa Skills Kit response
+# These utility functions construct the required JSON for a full Alexa Skills Kit response
+
+def build_response(session_attributes, speechlet_response):
+  return {
+    'version': '1.0',
+    'sessionAttributes': session_attributes,
+    'response': speechlet_response
+  }
+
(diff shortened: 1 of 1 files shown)


jingai:
Hold on merging this for a bit.  I think some extended error handling can make this better.

jingai:
@m0ngr31, starting work on this again tomorrow morning.  Is there any reason you chose to prefix the title and text with "SessionSpeechlet - "?  I'm wondering if that's something I need to preserve under some conditions?

jingai:
Reworked this a bit and merged the changes.  The logic is as described above, but also cleaned up the responses a bit.

Issue #22, opened 2016-10-05
Add certificate, URL, and AppID verification


jingai:
The skill should verify before continuing on:
- Application ID
- Incoming request is from Amazon
- Certificate URL comes from Amazon
- Timestamp of certificate is valid
- Certificate signature can be verified

I have code for all of this, but I need to factor it out of my other local changes.

jingai:
For AWS Lambda, I'm not in a position to test these changes, so I figured I'd talk a bit about them here.

To verify the certificate validity, I currently get at the certificate and signature via the server environment variables HTTP_SIGNATURECERTCHAINURL and HTTP_SIGNATURE.  I then check the DNS field is exactly 'echo-api.amazon.com'; the expiration time of the certificate is not greater than the current server time; the incoming request is from 's3.amazonaws.com' and the url starts with '/echo.api/'; and finally, that the signature actually verifies.

Does all of that sound like it would work on AWS Lambda?  I'm just trying to avoid people spamming you with support requests.

I'll push this all up to a branch for you to look at soon.

digiltd:
Sounds great if you are self hosting the Python side of things.

When using AWS Lambda I'm pretty sure all the security between Echo and the Skill is handled for you. 

https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/developing-an-alexa-skill-as-a-lambda-function

But of course that doesn't include the traffic in and out of your Kodi box.

jingai:
I honestly don't know, which is why I asked.  I can just only enable this stuff in the WSGI handler if that's what we decide to do.

digiltd:
According to the docs AWS does handle the security between the two

> - You do not need an SSL certificate.
> - You do not need to verify that requests are coming from the Alexa service yourself. Access to execute your function is controlled by permissions within AWS instead.
> - AWS Lambda runs your code only when you need it and scales with your usage, so there is no need to provision or continuously run servers.
> - Alexa encrypts its communications with Lambda utilizing TLS. 

Lambda uses the same wsgi.py so it would need to be a configurable option.

Do you know of a way we can refactor the code to accommodate these types of variations yet still keep the bulk of the kodi/alexa bit the same?  I have some local tweaks that I think would be worth sharing, i.e. using https and a reverse proxy to speak to your local network.

m0ngr31:
They both use the same `wsgi.py` file, but they are handled by different methods. 'lambda_handler' for lambda and 'application' for wsgi

[...]

Issue #21, opened 2016-10-05
WSGI: Store configuration in file


jingai:
Instead of environment variables.

I have code for this already, but I need to make it generic.

m0ngr31:
Can you update the main readme to show how to get a wsgi server running?

jingai:
Sure can, but I'm about to run from a hurricane, so it may take me a little bit lol.

m0ngr31:
No worries

jingai:
Regarding documentation, I'm not sure what the scope should be honestly.

Heroku uses WSGI, doesn't it?  I think that or something like it is what most will use, and you've documented that already.

For a self-hosted server, it depends on the web server the user chooses.. most are apache I'd guess, but it's enough to say, "use mod_wsgi" in that case, I think.  After all, if they can set up apache, they probably know that already?

digiltd:
> if they can set up apache, they probably know that already?

They might, but given the universal appeal of what this skill gives you and the popularity of Kodi, I expect many will be messing about without really knowing what they are doing. What are the numbers looking like for the traffic on this repo?

@m0ngr31 would you might turning on the wiki functionality for this repo?

Xaelias:
`they can set up apache, they probably know that already?`
Yes and know. I have an apache server running on a couple hosts because it comes with other services, or because I set it up to host a small static website. And clearly I have no idea what you just said :-)

You can never over explain something. You don't know who is going to come and use your tool. It does not hurt to explain something that might be obvious. Worst case scenario, people will skip that part.

jingai:
I mean that if someone doesn't know what they're doing, they're better off using one of the other solutions like Heroku or Lambda.  Hosting it yourself without general knowledge is dangerous, honestly.

Xaelias:
But what if somebody wants to learn?

jingai:
I get that, but that's why I asked what the scope is.  There is already plenty of documentation for setting up mod_wsgi out there.  How in-depth should we go for a README on an Alexa skill?

The WSGI handler is already documented here for Heroku.  From the standpoint of the skill, that's what you need to do for any other web server as well.

If I document here how to set up Apache from scratch, do I include nginx as well?  IIS?  Do I include how to properly load certs?

Xaelias:
If you know of good documentation, maybe just quote a couple links? I agree we can't explain everything, everywhere, all the time.
You obviously know more about these subjects that I do. So I'll honestly just trust you on this.

[...]

PR #32, opened 2016-10-14
Verify certificate and application ID

jingai:
In reference to Issues #21 and #22.

Adds two environment variables for the WSGI handler:
- KODI_APPID
- KODI_VERIFY_CERT

If KODI_APPID is set and not empty, the skill will verify that the incoming request appid matches before processing the request.

If KODI_VERIFY_CERT==1, the skill will perform various checks on the certificate:
- Incoming request is from Amazon,
- Certificate URL is from Amazon,
- Timestamp of certificate is valid/not expired,
- Signature can be verified.

Additionally, this PR adds the ability to store all configuration in a file, 'wsgi.ini', which can look like:

```
[general]
app_id = amzn1.ask.skill.deadbeef-4e4f-ad61-fe42-aee7d2de083d
verify_cert = yes

[kodi]
address = 127.0.0.1
port = 8080
username = kodi
password = kodi
```

If we want to extend this to the Lambda handler, I'm happy to do that.  I just wasn't comfortable doing it since I don't know much about it and cannot test it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,4 @@ test.py
 .vscode
 
 .env
+*.ini
--- a/kodi.py
+++ b/kodi.py
@@ -89,8 +89,8 @@ def SendCommand(command):
   USER = os.getenv('KODI_USERNAME', 'kodi')
   PASS = os.getenv('KODI_PASSWORD', 'kodi')
 
-  print KODI
-  
+  print "Sending request to %s:%d" % (KODI, PORT)
+
   url = "http://%s:%d/jsonrpc" % (KODI, PORT)
 
   try:
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,7 @@ requests
 gunicorn
 pycountry
 ConfigParser
+aniso8601
+six
+OpenSSL
 yaep
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,5 @@
 requests
 gunicorn
+yaep
 pycountry
 ConfigParser
-aniso8601
-six
-OpenSSL
-yaep
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 requests
 gunicorn
 pycountry
+ConfigParser
 yaep
--- a/requirements_verify.txt
+++ b/requirements_verify.txt
@@ -1,0 +1,3 @@
+aniso8601
+six
+OpenSSL
--- a/verifier.py
+++ b/verifier.py
@@ -1,0 +1,67 @@
+import os
+import base64
+import posixpath
+from datetime import datetime
+from six.moves.urllib.parse import urlparse
+from six.moves.urllib.request import urlopen
+
+from OpenSSL import crypto
+
+
+class VerificationError(Exception): pass
+
+
+def load_certificate(cert_url):
+    if not _valid_certificate_url(cert_url):
+        raise VerificationError("Certificate URL verification failed")
+    cert_data = urlopen(cert_url).read()
+    cert = crypto.load_certificate(crypto.FILETYPE_PEM, cert_data)
+    if not _valid_certificate(cert):
+        raise VerificationError("Certificate verification failed")
+    return cert
+
+
+def verify_signature(cert, signature, signed_data):
+    try:
+        signature = base64.b64decode(signature)
+        crypto.verify(cert, signature, signed_data, 'sha1')
+    except crypto.Error as e:
+        raise VerificationError(e)
+
+
+def verify_timestamp(timestamp):
+    dt = datetime.utcnow() - timestamp.replace(tzinfo=None)
+    if dt.seconds > 150:
+        raise VerificationError("Timestamp verification failed")
+
+
+def verify_application_id(candidate, records):
+    if candidate not in records:
+        raise VerificationError("Application ID verification failed")
+
+
+def _valid_certificate_url(cert_url):
+    parsed_url = urlparse(cert_url)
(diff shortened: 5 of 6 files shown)


jingai:
Could someone who's using Heroku test with this please?

Specifically, I'd like to make sure it doesn't break existing installations -- that is, try dropping this in with no configuration changes.

Second, migrate to the wsgi.ini configuration, without adding the app_id and verify_cert parameters.

Finally, add your app_id and enable verify_cert.

If any of those make it fail, please report here.

m0ngr31:
Just from looking at the code, I think it should be fine. I just wonder what's the benefit of the .ini file? Environment vars are great for keeping secret info out of version control. I guess if you are running this locally, you aren't going to be pushing anything to the cloud...

jingai:
I tried to (and think I did) retain the ability to use environment variables.  In fact, the default (no ini file) shouldn't interrupt anyone's existing installations.

I personally keep wsgi.ini 'secret' by adding it to .gitignore (which is included this PR) and an Apache directive to ignore the file.

The primary purpose though is that I host this skill for a few family members, so I have a few copies of it on the same server.  With an ini file, I just clone to a new directory and modify the wsgi.ini in there to suit them.

If you don't run each one as a separate instance (in a VM or whatever), you have namespace conflicts since the environment variables are global.

m0ngr31:
That makes sense. I saw that you allowed the env vars still, I was just wondering about the .ini purpose, but it sounds perfect for your use case.

jingai:
I realize it's a very specific use-case that probably won't be so widely-used these days, but at least at the moment it doesn't seem like a terrible burden to maintain both paths.

Your repo and your call, though.  After all, that's why I wanted to consult you on this one first :)

jingai:
@m0ngr31, if you want to try adding the cert/appid verifier to the Lambda handler it'd be pretty trivial to adapt this to it.  Just add KODI_APP_ID and KODI_VERIFY_CERT environment variables to .env and factor out the verification code from wsgi_main_handler so you can call it from lambda_handler.

I have no idea if it's desirable or necessary to do so, though.

Alternatively, after/if we merge this, I can make another branch that does this for you to test.

jingai:
@m0ngr31 , is it OK if I merge this tomorrow (Oct 21) if we don't get any responses from testers?  I have tested the various conditions here (no wsgi.ini, no environment variables, a mix, etc).

I'll then make a branch with the verifier enabled for Lambda that you can test if you like.

digiltd:
If there is a way to conditionally install the added cert dependencies (ConfigParser, aniso8601, six, OpenSSL) for a self hosted install, that would be peachy.

jingai:
Do you mean to provide an option to not import those?  Or are you asking for the skill to install the dependencies automatically?

jingai:
BTW, I'll document this after we decide whether or not to include it in the Lambda handler.  If it works, I'm inclined to do so, as even if it's not totally necessary, it is still optional (and off by default).

digiltd:
AWS Lambda uses PIP and installs everything in the `requirements.txt` file when you deploy. The   package size goes from 15.7mb to 27mb with these extras. Each time you deploy using the lambda-deploy method, AWS versions the whole thing so the size your "code storage" increases by 27mb. It doesn't take long before you are using hundreds of MB

I am not entirely sure if this is what is supposed to happen or if there is a different way to deploy. The idea of uploading and versioning dependencies when you change a single variable name doesn't seem very efficient. I am looking into it. But currently I am deleting the function from AWS every now and then to keep the code storage used low.

The memory used to import the unused modules should be considered, I am not sure if it affects performance but the advice on the AWS docs say:

> Lower costs and improve performance by minimizing the use of startup code not directly related to processing the current event.

Mine currently uses around 50 of the 128 mb allocated when searching.

Though given this Skill isn't prime time it doesn't need to be super efficient, but every little helps :)

Speaking of which, i have just tried not importing pycountry due to the package being 13mb. It reduced my execution times by 1 second, code storage is 1.7mb and memory usage when searching 27mb. I will start another issue but thought I would mention it.

jingai:
We can talk about potential optimizations when we're a bit further along.

For now, if you've got some limit to consider, I'd just keep deleting old versions from N on, where N is whatever you deem reasonable.

For pycountry, I can wrap the import with a try/except/pass, and do the same in the one spot where it calls it, but I am not sure how to prevent it from installing it.  That is, can you have conditional statements in requirements.txt?  And what would the condition be?

But you're right, this should all be in another Issue :)

m0ngr31:
Could we have a separate file with the requirements for WSGI?

So people on Lambda could have the stripped down version, and people who want the extra security could do a `pip install -r wsgi.txt` or something?

jingai:
Sure, I can wrap the cert dependencies in a try/except/pass and move them from requirements.txt to.. say.. requirements_extras.txt?

Should I move the pycountry dependency there too?  I thought it was kind of cool to have Alexa speak the language when you're switching subtitles, but I do agree it's a bit heavy.  I just couldn't find any other way to translate the ISO 3166-1 alpha-3 codes into English names.

jingai:
I've moved aniso8601, six, and OpenSSL to requirements_verify.txt.

ConfigParser is still there as an unconditional dependency because I'd like to factor it out to support Lambda deployments too, as it simplifies documentation should someone try it.

pycountry is still unconditional as well, but we can revisit that later if it's found to be a problem.

I'll follow this up with two more PRs for Lambda:
- wsgi.ini support
- cert/appid verifier

After the above are accepted or denied, I'll update the documentation to reflect the changes.

PR #34, opened 2016-10-18
Updated REGION variable so deploy would work

irvingwa:
Changed REGION variable to AWS_DEFAULT_REGION.


jingai:
That's a lot of commits for a one-line change.  I've gone ahead and just made a single commit for this change with 8b8e0eb2f7c86892250cf737ff03b6386c7c4836

m0ngr31:
I'll make this change with PR #37

PR #35, opened 2016-10-18
add a new env variable to define the SCHEME

digiltd:
I am not sure how this affects Heroku, there is a default to HTTP which should prevent any breakages due to the KODI_SCHEME env variable not being set.

Would the only Heroku change be the setting of the env var by adding it to the `config:set` command i.e. 

`heroku config:set KODI_SCHEME='https' KODI_ADDRESS='your_ip_or_dynamic_address' KODI_PORT='kodi_port' KODI_USERNAME='kodi_username' KODI_PASSWORD='kodi_password' --app app-name-and-number`.

If that is the case I will change the readme to include it.

### DIFF
--- a/.env.lambda
+++ b/.env.lambda
@@ -1,15 +1,29 @@
-SKILL_APPID =
-KODI_ADDRESS =
-KODI_PORT =
-KODI_USERNAME =
-KODI_PASSWORD =
-LAMBDA_ENV_VARS = KODI_PASSWORD,KODI_USERNAME,KODI_PORT,KODI_ADDRESS,SKILL_APPID
+
+KODI_ADDRESS = 
+KODI_PORT = 
+KODI_USERNAME = 
+KODI_PASSWORD = 
+
+# The Kodi webserver only supports HTTP. 
+# Uncomment KODI_SCHEME to tell the skill to use https between AWS and your local network
+# (only use if you have already set this up with your own certificates) 
+
+# KODI_SCHEME = https
+
+# If using a reverse proxy you might need to add an extra bit to the url before "jsonrpc"
+# You can do this with KODI_SUBPATH (don't use slashes before or after) 
+
+# KODI_SUBPATH = 
+
+SKILL_APPID = 
+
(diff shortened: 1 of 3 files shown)


jingai:
Stupid question.. but where do I even enable HTTPS in Kodi?  http://kodi.wiki/view/JSON-RPC_API#HTTP doesn't list it as an option, nor do I see it in Kodi's configuration.

digiltd:
You don't :)

But if you are not running the skill locally then the KODI_ADDRESS var would need to be your dyndns address and that can be HTTPS e.g. `https://dyndns.domainname.com:33133`

Then you port forward or use a reverse proxy to direct traffic to your Kodi local ip using HTTP.

jingai:
Ah I gotcha.

Looks fine to me then.

digiltd:
Squashed commits

Also updated readme to include the KODI_SCHEME var in the Heroku command and added information about what should go into the vars.

jingai:
It'd be relevant to this PR if you want to add the variable for specifying the path too (KODI_JSONRPC_PATH?  Or maybe just KODI_PATH, since KODI_SCHEME doesn't describe what it's modifying either).

digiltd:
I could, but as this PR is (hopefully) ready to merge it would be good to tick it off and I can clean up  a bit.

jingai:
Well, it isn't necessary, but this PR won't work if the user can't proxy on the default path for whatever reason (e.g., has another configured already).

Also, you should probably add something to the documentation describing that this is for a proxy.  It wasn't initially clear to me and I'm going to guess other users are just going to assume that Kodi supports HTTPS now and ask about it on the forums.

digiltd:
Added env vars for KODI_SCHEME and KODI_SUBPATH and included some comments in the .env files.

Tested with AWS only, defaults are in place to not upset any installs that don't use the extra vars.

jingai:
Looks OK at a glance.  I'll make sure there's no issue with the defaults when I get home today.

jingai:
Sorry, I'm not ignoring this -- just have been coming home so late I don't have the energy to get to it.  Should be over with soon though.

jingai:
All merged.  Removed some trailing whitespace and shuffled a bit around, but it's the same functionally.

PR #36, opened 2016-10-18
int definition for PORT changed to string

digiltd:
This allows the use of reverse proxy rules to be based on a subfolder in the url

e.g. `http://dyndns.domainname.com:33133/dosomethingiftheaddresshasthissubfolder`

Resulting in the full jsonrpc concatenated url:

`http://dyndns.domainname.com:33133/dosomethingiftheaddresshasthissubfolder/jsonrpc/request?=`

I am making these small tweaks to accompany a little guide I am putting together about how to set up a secure connection to a reverse proxy that only accepts traffic from a specific group of Amazon IPs.

I don't want to get people messing in the code to make these tweaks themselves, and with the change in the PR you can set the subfolder in the KODI_PORT var 

e.g. `KODI_PORT = 33133/dosomethingiftheaddresshasthissubfolder`

### DIFF
--- a/kodi.py
+++ b/kodi.py
@@ -85,13 +85,13 @@ def SetupEnvVars():
 def SendCommand(command):
   # Change this to the IP address of your Kodi server or always pass in an address
   KODI = os.getenv('KODI_ADDRESS', '127.0.0.1')
-  PORT = int(os.getenv('KODI_PORT', 8080))
+  PORT = os.getenv('KODI_PORT', '8080')
   USER = os.getenv('KODI_USERNAME', 'kodi')
   PASS = os.getenv('KODI_PASSWORD', 'kodi')
 
   print KODI
   
-  url = "http://%s:%d/jsonrpc" % (KODI, PORT)
+  url = "http://%s:%s/jsonrpc" % (KODI, PORT)
   try:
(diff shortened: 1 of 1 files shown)


jingai:
Instead of overloading KODI_PORT, wouldn't making a new environment variable be better/clearer?

m0ngr31:
Should probably update the Slot generator code for this as well.

digiltd:
I agree it would be cleaner, though I thought it would be messier having to explain the extra env var and add the logic to handle it.

I was also being a little lazy as KODI_PORT is only used once to concat the url and making it string wouldn't cause any issues.

jingai:
Well, they're all only used once currently really.  I'd personally make it a new variable since, at the least, the value doesn't match the description of the key in the latter use; i.e., '33133/dosomethingiftheaddresshasthissubfolder' is definitely not a 'port'.

jingai:
You don't need to close the PR to re-do the branch, btw.  You can either a) force-push to the branch, or b) delete the old branch and then push another with the same name.  Github figures it out :)

Pushing continually to a branch during the discussion of a PR is a pretty common thing.  Generally it's just expected that you'll rebase -i when you're done to squash them down, then re-push the branch.

digiltd:
Thanks for the advice, it is appreciated :)

PR #45, opened 2016-10-26
Add the capability to resume movies and TV-Shows

Xaelias:
I'm often in the situation where I stop an episode or a movie in the middle. This usually happens when I start an episode over lunch and don't have time to finish it.

Thing is, there is currently no possibility to ask kodi to just keep going wherever I last stopped.

This commit allows just that. If the user asks to play a movie, kodi will resume if it was started earlier. If the user asks to play the next episode of a given show, kodi will either resume the last episode if it was not entirely watched, or play the next episode if available (i.e. what happens today).

Exceptions:
- random movies and episodes
- asking to play a specific episode from a TV show

Let me know how that sounds :-)

I also took the liberty to remove the functions used to manage the video playlist since they were not used anymore.

### DIFF
--- a/kodi.py
+++ b/kodi.py
@@ -192,6 +192,12 @@ def PrepMoviePlaylist(movie_id):
 def StartVideoPlaylist():
   return SendCommand(RPCString("Player.Open", {"item": {"playlistid": 1}}))
 
+def PlayEpisode(ep_id, resume=True):
+  return SendCommand(RPCString("Player.Open", {"item": {"episodeid": ep_id}, "options": {"resume": resume}}))
+
+def PlayMovie(movie_id, resume=True):
+  return SendCommand(RPCString("Player.Open", {"item": {"movieid": movie_id}, "options": {"resume": resume}}))
+
 def AddSongsToPlaylist(song_ids):
   songs_array = []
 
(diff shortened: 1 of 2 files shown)


jingai:
I'm not sure we want to remove the video playlist functions.  They weren't ever used AFAIK, but I do want to use them in the future for, e.g., music videos.  For episodes and movies, it doesn't hurt to leave them in for later if we think of a use for them (e.g., "Alexa, play all of season 2 of {show}").

Xaelias:
Ah work habits. At my job we never leave unused code in the code-base. Even if we believe we might still use it in the future, we know it's somewhere in git history.
Anyway, I can add them back I don't really care :-)

I was also thinking, for the movies, I should probably add a resume target instead of assuming resuming is always the default action. "ask kodi to play movie" and "ask kodi to resume movie". Thoughts?

jagsta:
I think having a resume intent would be good to provide the choice of which you'd like to happen.

Alternatively, it might be good to have an interaction, so if the movie/episode is resumable to be asked if you'd like to resume or play from the beginning. I'm not sure how difficult this would be to achieve, I haven't looked into multi stage conversations with Alexa yet.

Xaelias:
Yeah me neither. I honestly don't see the use for episodes. I don't think I ever stopped an episode in the middle, came back 2 months later and wanted to start it over again.

jingai:
@Xaelias, two reasons I'd rather you leave the playlist stuff in: a) I plan to use it soon, and b) it's not really relevant to this PR :)

Leaving the session open and having Alexa ask if you'd like to resume or restart would be ideal, but we don't yet have the foundation for that.  It's one of the things I'd planned on looking at soonish, though.

For now, my suggestion is to leave it as you have it now, because you can always say "Alexa, ask Kodi to start over" if you didn't want it to resume.

Xaelias:
It's relevant because I removed all the calls to these functions. But yeah I'll put them back in. I'll update the PR later today no problem :-)

jingai:
Just a comment about resuming episodes too: there is one good use I can think of for it that my family would make use of.. my wife and I get interrupted when watching shows a lot by our kids that want us to put on some cartoon instead, but we'll generally resume the episode within the same day.

Xaelias:
Kids... :p Are you responding to my second to last comment? What I meant by I don't the use for episodes was I don't see the use of 2 different utterances.
Like having "play the next episode of Supernatural" and "resume Supernatural" or something like that. It's unlikely, in my mind, to have an episode that I started to watch, and want to start it over. If that happens, I can go the extra mile of asking kodi to start over like you mentioned above.

But yeah, I definitely will use the resume for TV Shows on a weekly if not daily basis. Honestly I'm surprised nobody added this functionality earlier.

jingai:
Yup, sorry, just misunderstood what you said earlier.  But yeah, we're on the same page now.

I'd say this is good for merge after you add the playlist code back in and squash.  I'll test as soon as I can -- have been coming home at 11-midnight from work lately, so it might be a day or two.

Xaelias:
Hey no hurry. I deployed it on mine yesterday. Just thought I would share :-) I probably won't get to update the PR until tonight anyway.

Xaelias:
PR updated :-)

jingai:
Does the "resume" option not work when calling Player.Open with a playlist?

edit: Couldn't sleep (oh how I wish I'd gotten a flu shot earlier...) so I tried this, and it doesn't.  Guessing the resume option works on the playlist position rather than the player position in this case.

jingai:
Thanks for the submission!  All merged (despite what github thinks).

Added a bit in another commit too to speak whether it's Resuming or Playing :)

Xaelias:
@jingai Not that I could find in the API (for the playlist resume). The only thing that could work I think would be to store the resume point, and manually seek after starting the playlist. Which would work for the first episode, but is really messy in mo opinion. I just don't see the benefit of using the playlist in this case.
Has for the unmerged commits, it's probably because I forced pushed earlier. Might have messed with github algorithms.

jingai:
It's no problem either way, was more curious than anything.  Thanks again!

PR #46, opened 2016-10-26
Creating a Python version of PR #28

m0ngr31:


jingai:
Two things stand out:
1. Should it copy the default .env.lambda for deployment if .env doesn't exist?  It won't work without editing first, so that might just be more confusing than a hard error,
2. Should it remove the temporary directory it created for deployment?

m0ngr31:
You're probably right on 1. I can blast that real quick. Hopefully that'll remove any confusion.

For 2, I figured the safest way to always have the freshest code is to just blast it and re-copy the files over.

jingai:
I'm indifferent on 2, partly because I don't use it anyway ;)

Otherwise seems OK.  @digiltd, mind giving it a test?

digiltd:
Just tested it and it worked.

Mac users who installed Python with Homebrew (like me) have to include a setup.cfg file containing:

```
[install]
prefix= 
```

For the PIP installation part of the deploy. This is the first time I have had to include the prefix, usually PIP works fine without it. So is it worth adding

``` python
if(os.path.isfile("setup.cfg")):
  shutil.copy("setup.cfg", app_name)
```

Or do we assume that if they have to use the `setup.cfg` they will know what to do with it? I have been using Homebrew Python for years and this is the first time I have used it.

jingai:
You could always generate app_name/setup.cfg in the script, to avoid having a weird file sitting in the repo.

m0ngr31:
I've never heard of this before, but I'm on Linux. I'll look into it when I get a minute.

digiltd:
Yeah, I think I will leave it then. Though it would be a helpful bit of info that could go on a wiki.

digiltd:
@m0ngr31 It doesn't break anything but if you are interested here is the error displayed when running `python deploy-to-lambda.py` without it:

```
Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/local/lib/python2.7/site-packages/pip/commands/install.py", line 317, in run
    prefix=options.prefix_path,
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_set.py", line 742, in install
    **kwargs
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_install.py", line 831, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/usr/local/lib/python2.7/site-packages/pip/req/req_install.py", line 1032, in move_wheel_files
    isolated=self.isolated,
  File "/usr/local/lib/python2.7/site-packages/pip/wheel.py", line 247, in move_wheel_files
    prefix=prefix,
  File "/usr/local/lib/python2.7/site-packages/pip/locations.py", line 153, in distutils_scheme
    i.finalize_options()
  File "/usr/local/Cellar/python/2.7.12_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/command/install.py", line 264, in finalize_options
    "must supply either home or prefix/exec-prefix -- not both"
DistutilsOptionError: must supply either home or prefix/exec-prefix -- not both
A DistutilsOptionError about the prefix can occur when you are on OS X and installed Python via Homebrew.
If this is you, please look at https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/Homebrew-and-Python.md#note-on-pip-install---user
If this is not you, please contact us  for support.
```

As you can see the error highlights that the problem might be Homebrew related and even gives a link for Homebrew users to follow.

digiltd:
Perhaps add to the comment that it includes the .git folder.

> By default lambda-deploy will package the entire folder it is run from (including the .git folder) and upload it to AWS each time you deploy. To keep things a little neater you can run this script which will copy the Lambda relevant files to a separate folder then deploy them from the new folder.

m0ngr31:
From what I can see, it should be a profile setting right? So the file would be at `~/.pydistutils.cfg`, and not in the actual deploy folder? Or am I missing something?

digiltd:
It isn't something that I have had to do before when using pip. Just to clarify, installing on the Mac is fine using the usual `pip install -r  requirements.txt`. Only when lambda-deploy tries to deploy it does it become an issue.

I wouldn't worry about it, the error message from pip should be enough of a hint if it happens to someone else.

jingai:
One aspect to consider is that it's likely to generate support requests (i.e., forum posts).  Pretty sure @m0ngr31 is right: this is a dev environment issue, not a skill issue.

But it's still worth thinking about.

m0ngr31:
I can add something to the README if it becomes a problem

digiltd:
@jingai ish, I would say it is a deployment issue. If following the instructions to deploy to AWS you need to have pip installed. 

Though just by having it written down in this thread for others to find is already very helpful.

PR #47, opened 2016-10-26
New intents - 'watch PVR channel x' and 'switch to whatever PVR channel is broadcasting x'

freemans13:


jingai:
Thanks for cleaning it up a bit.

It might be a while before I can review this though, as my real job is kicking my butt at the moment, and I don't have a PVR to test any of it.  So I'd want to have the time to sift through it carefully.

@m0ngr31 may get a chance sooner, but I can't speak for him.

Xaelias:
Don't have a PVR either. But I can try to look at the change anyway.

alexa.utterances
```diff
@@ -648,3 +648,25 @@ WhatNewShows what shows do you have
...
+WatchPVRBroadcast find the broadcast called {Broadcast}
+WatchPVRBroadcast find {Broadcast}
+WatchPVRBroadcast watch something called {Broadcast}
+WatchPVRBroadcast watch anything called {Broadcast}
+WatchPVRBroadcast jump to {Broadcast}
+WatchPVRBroadcast change channel to {Broadcast}
+WatchPVRBroadcast change channel to watch {Broadcast}
+WatchPVRBroadcast change channel to play {Broadcast}
+WatchPVRChannel watch {Channel}
+WatchPVRChannel play {Channel}
```
jingai: Watch and Play need a modifier, at least for now.

I want to make it so the skill can cope with "play foo", but at least currently, "play foo" is mapped to music artists.

Ideally, "play foo" would attempt to search through all available media and do smart things, but it's not there yet.

alexa.utterances
```diff
@@ -648,3 +648,25 @@ WhatNewShows what shows do you have
...
+WatchPVRBroadcast play the broadcast called {Broadcast}
+WatchPVRBroadcast look for the program called {Broadcast}
+WatchPVRBroadcast look for the broadcast called {Broadcast}
+WatchPVRBroadcast look for {Broadcast}
+WatchPVRBroadcast where is the program called {Broadcast}
+WatchPVRBroadcast where is the broadcast called {Broadcast}
+WatchPVRBroadcast where is {Broadcast}
+WatchPVRBroadcast find the program called {Broadcast}
+WatchPVRBroadcast find the broadcast called {Broadcast}
+WatchPVRBroadcast find {Broadcast}
```
jingai, line 1943: "look for {Broadcast}," "where is {Broadcast}," and "find {Broadcast}" shouldn't yet exist.

Currently, by default (without qualifiers on the media type), music artist is assumed when no type is specified.

This will change eventually, but for now, all other media types need to be specified in the command.

lukesmith:
I've done some testing and `WatchPVRBroadcast` and `WatchPVRChannel` both work as I would expect. Though I find using `watch` gets Alexa confused, thinking I'm saying `what`.

~~Also I don't find `WatchPVRBroadcast` too accurate. I assume having a process to keep the `BROADCASTS` custom slot regularly updated would improve the accuracy of this?~~

Just discovered the `generate_custom_slots` script. Working better now

freemans13:
Sorry for the delayed response. Work ties my up from time to time.

I'd love some suggestions for the 'media type' and the utterances we'd like to see in place. 
Is the media type PVR, DVR, TV, Live-TV or something else?

For utterances I can see we use this to watch a TV channel:
WatchPVRChannel (play/watch) PVR {Channel}

I'm struggling for an utterance for other ability where it switches to the channel that is currently (or will be) showing {Broadcast}. Alexa gets confused with Broadcast and Program for me.
Any ideas much appreciated!

jingai:
I'll give it some thought as soon as I have some free time.

Just FYI though: instead of ```git merge master``` you can ```git rebase master``` to pull upstream changes back into your branch.  This avoids all the "merged master" commits.

freemans13:
Thanks for the git tip. I'm new to github pull request management as you can tell.

devdems:
I tested it with tvheadend pvr on kodi and it works fine.

Would be nice to add something like "play channel {channel number}".

devdems:
I have now tested this for a week. We should merge this pull request because pvr is one of the key things that kodi is used for.

m0ngr31:
Unfortunately, it's going to have to wait on the other branch. And then it will need refactoring before it can be merged in.

jingai:
We will need to look over the utterances added here carefully too.  Some of them I think shouldn't exist (yet), and some probably should be part of pre-existing intents (the 'find', 'where is', etc, utterances likely should be part of DoSearch, for instance).

devdems:
Is this going to be implemented?

jingai:
Eventually, I'm sure it will be, but if you read the above comments, you'll see there's still a fair amount left to do.

I don't have a PVR so it's hard for me personally to justify spending the time required on it, unfortunately.

devdems:
Let us know if there is anything we can do to speed up this. As there are a huge amount of people that use pvr in kodi it will be easy to find testers.

m0ngr31:
I'm going to just close this and if you make a compatible version in the future, just open a new PR.

A suggestion for the next version: Channels are going to be better off as slots and trying to match them the current way before resorting to what you have here.

PR #54, opened 2016-11-06
Fuzzy matching in matchHeard

jagsta:
I've been having some problems matching a few of the titles in my libraries, these have mostly broken down into four problem areas:

1. Titles/Artists with numbers as words (Zero Dark Thirty, Zero 7, Four Lions, The Hateful Eight)
2. Titles which sound like other words which Alexa interprets incorrectly (Samsara -> Some Sarah)
3. Titles with abbreviations (Dr. Strangelove)
4. Alexa mishearing what I asked for.

I've been using a couple of approaches to help reduce failures to match the above with reasonable success.

1. Fuzzy matching using the fuzzywuzzy python module (and the python-Levenshtein module to speed up the execution of fuzzy matches)
2. Replacing digits in Alexa's heard request with words.

I've made some additions to the matchHeard function to provide a 3rd tier of matching attempts when the existing two methods fail to find a match, along with two helper functions which are used to replace numbers with words.

The word_form function is mainly taken from a stack overflow answer, I didn't have the heart to strip out the higher levels it's capable of converting to words, although they'll almost certainly never be used. It's been modified to return the word zero as it didn't previously.

I've read the new contributing file, but haven't updated the changelog as I don't know what the versioning standard is and haven't added any features as such in this PR.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,5 @@ requests
 gunicorn
 yaep
 pycountry
+python-Levenshtein
+fuzzywuzzy
--- a/kodi.py
+++ b/kodi.py
@@ -37,6 +37,9 @@ import re
 import string
 import sys
 import pycountry
+from fuzzywuzzy import fuzz
+from fuzzywuzzy import process
+
 
 # These are words that we ignore when doing a non-exact match on show names
 STOPWORDS = [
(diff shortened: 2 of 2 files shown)


jingai:
Great stuff.  Will have give it a test run as soon as I get a chance!

requirements.txt
```diff
@@ -2,3 +2,5 @@ requests
 gunicorn
 yaep
 pycountry
+python-Levenshtein
```
jingai, line 5: Is this required here?  Doesn't fuzzywuzzy require it already?

edit: Nevermind!  I see it's an optional requirement.

jagsta: It's listed as an optional dependency in the module documentation. Looking at the code it seems to try and import Levenshtein based sequence matching functions and fall back to difflib if that fails:

https://github.com/seatgeek/fuzzywuzzy/blob/master/fuzzywuzzy/fuzz.py

jingai: Yup, sorry for the noise.

jingai:
Excellent work @jagsta , this really helps!

Cleaned up some tab stops to match the rest (yeah, sorry.. can't help myself), but otherwise merged as-is 👍

jagsta:
Great! It could doubtless be improved further but it's already made a difference to my success rates for difficult titles. The thresholds and the specific matching methods used might benefit from being modifiable via environment variables if success rates and false positives vary with other peoples libraries.

jingai:
If people report issues, we can break it out as an option.  I tried all of the 'difficult' titles though and it worked quite well.

The only issue I found remaining is titles with roman numerals.  For instance, saying "Star Wars Episode Six" matched "Star Wars Episode Four".  This isn't a problem specific to your code, though.

PR #55, opened 2016-11-09
Changes to WhatNewShows intent

jingai:
Making this a PR because it changes some existing behaviors.

In accordance with the 5-movie-response limit of the WhatNewMovies intent, I've restricted WhatNewShows to the same.  I'm open to changing the default value, but did feel it needed a limit for those of us with very large libraries.  This goes hand-in-hand with the next change too..

..I've removed the dateadded constraint on the list of shows to consider.  Previously, it would only consider episodes added in the last 5 days.  This is reasonable perhaps if the user views shows on Kodi in manor similar to cable TV, but I suspect that's not often the case.  Given the sheer number of shows my family watches, we frequently won't get around to an episode for quite some time, but I still want Alexa to remind us of it.

Lastly, I've merged CheckNewShows intent into WhatNewShows.  They simply didn't seem dissimilar enough to have both.

All of these changes are broken out into separate commits to make it easy to back one out if desired.

### DIFF
--- a/alexa.intents
+++ b/alexa.intents
@@ -1,10 +1,6 @@
 {
   "intents": [
     {
-      "intent": "CheckNewShows",
-      "slots": []
-    },
-    {
       "intent": "WhatNewMovies",
       "slots": []
     },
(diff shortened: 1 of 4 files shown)


m0ngr31:
Looks great, thanks!

PR #57, opened 2016-11-10
Rename step commands

digiltd:
Changed "step" to "seek" and prepended them all with "Player" for better relationship with the Kodi methods.

Initially it was just keep them grouped in the utterances file when that is sorted alphabetically

The commands are still "step forward" for now until seek is working with numerical values (in development)

### DIFF
--- a/alexa.utterances
+++ b/alexa.utterances
@@ -12,50 +12,14 @@ AudioStreamPrevious previous audio stream
 AudioStreamPrevious previous language
 Back go back
 Back navigate
-BigStepBackward big back jump
-BigStepBackward big back step
-BigStepBackward big backward jump
-BigStepBackward big backward step
-BigStepBackward big jump back
-BigStepBackward big jump backward
-BigStepBackward big step back
-BigStepBackward big step backward
-BigStepBackward large back jump
-BigStepBackward large back step
-BigStepBackward large backward jump
-BigStepBackward large backward step
-BigStepBackward large jump back
-BigStepBackward large jump backward
-BigStepBackward large step back
-BigStepBackward large step backward
-BigStepBackward long back jump
-BigStepBackward long back step
-BigStepBackward long backward jump
-BigStepBackward long backward step
-BigStepBackward long jump back
-BigStepBackward long jump backward
-BigStepBackward long step back
-BigStepBackward long step backward
-BigStepForward big forward jump
-BigStepForward big forward step
-BigStepForward big jump forward
-BigStepForward big step forward
-BigStepForward large forward jump
-BigStepForward large forward step
-BigStepForward large jump forward
-BigStepForward large step forward
-BigStepForward long forward jump
-BigStepForward long forward step
-BigStepForward long jump forward
-BigStepForward long step forward
 CleanAudio clean audio library
 CleanAudio clean music library
 CleanVideo clean library
 CleanVideo clean video library
 ContinueShow continue last show
-ContinueShow continue show
 ContinueShow continue playing last show
 ContinueShow continue playing show
+ContinueShow continue show
 ContinueShow continue watching last show
 ContinueShow continue watching show
 CurrentPlayItemInquiry what audio is currently playing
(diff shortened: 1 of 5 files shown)


alexa.utterances
```diff
@@ -525,6 +514,41 @@ PlayerZoomOutMoveUp zoom out and pan up
...
+PlayRandomEpisode watch random episode of {Show}
+PlayRandomMovie play a random film
+PlayRandomMovie play a random movie
+PlayRandomMovie play random film
+PlayRandomMovie play random movie
+PlayRandomMovie watch a random film
+PlayRandomMovie watch a random movie
+PlayRandomMovie watch random film
+PlayRandomMovie watch random movie
 Prev listen to previous
```
jingai, line 568: Why did you move these?  This file was recently auto-generated, and the next time it is, they'll just move back anyway.

edit: I'm guessing you used a sort tool that sorts uppercase below lowercase, but the expansion tool we use at http://www.makermusings.com/amazon-echo-utterance-expander/ produces results in line with GNU sort; i.e., uppercase is sorted above lowercase.

CHANGELOG.txt
```diff
@@ -1,3 +1,6 @@
+v2.0.6 (11/10/2016)
+* Changed Step intents to PlayerSeek for easier grouping, functionality unchanged.
+
```
jingai: I don't think we need to release a new version for a non-functional change.  It probably doesn't even need a mention in the changelog, honestly.

jingai:
This PR contains two commits that claim to do the same thing -- mistake?

jingai:
Rebase and squash if you would please, and I'll pull this in.

digiltd:
I think there should be a mention in the changelog. It is a change, and it might be helpful to someone at some point.

I have changed the changelog and put the change under **unreleased**, I guess this can be used for minor pulls etc until something worthy of a release comes along.

jingai:
The git repo contains history in and of itself, though.  I don't think we need to outline every behind-the-scenes change in the CHANGELOG, but rather use that for user-facing changes/additions to functionality.

digiltd:
True, though I know I find it helpful to see all the changes without having to dive into commits. If we followed a similar template to https://github.com/olivierlacan/keep-a-changelog/blob/master/CHANGELOG.md it is easy to identify the type of changes.

I know that I don't always bother updating the intent schema and utterances for each new commit on the master branch. With this change if you don't then it doesn't work.  

My thinking is that it's more helpful to include a change than it is to leave it out.

jingai:
'git log' isn't any harder than 'cat CHANGELOG' though, and I don't think it's worthwhile to duplicate them.  In fact, for summaries, I tend to structure my commit comments like they do in the Linux kernel:  the first line is a one-line summary of the change, and what follows is a more in-depth explanation/rationalization of the change.

I get that your rationale is that it involves a change to the user's copy of the skill, but my thinking is more that the user should _plan_ on this for every pull.  Due to the fact that we don't host the skill for the user, it's just part of the update process.  The two must be kept in sync.  For those that are annoyed with this or don't want to deal with potentially disastrous bugs, etc, sticking to tagged releases is probably the best option.  (edit: in fact, modifying the README to direct users to download the ZIP release is probably a good idea -- those that want to and know what it entails can pull whatever branch they want, but this shouldn't be the default instruction)

Lastly, I don't include every commit in the changelog primarily because it wasn't done that way before, and I've learned to try to stick to the existing codebase's style and practices when contributing to someone else's project.  If you look at the changelog in its current state, there's a lot of underlying code changes that are not mentioned there.  It's all user-facing changes and additions.

There's no formal style guidelines or anything here, but this is my opinion:  changelogs are something users read, and the commit history is something developers read.

digiltd:
I don't know what to suggest for the best then.  The contents of a changelog are subjective and my opinion does differs from yours on the matter :)

I'm all for respecting the codebase and [contributing guidelines](https://github.com/m0ngr31/kodi-alexa/blob/master/CONTRIBUTING.md), they are the reason I added anything to the changelog at all :)

jingai:
I can leave this for @m0ngr31 to decide then. 

I do feel that if you want to add one line item like this to the changelog, you should go back through the commit history and add the others. 

I think you'll find it makes the changelog rather wordy though.

jingai:
Merged, but we should probably iron out the changelog policy.  Here is as good a place for discussion as any since it's already going on here.

@m0ngr31, one other thing I wanted to bring up is:  do we really want a version bump for each and every feature/change?  I suspect you added that bit to the guidelines just to avoid having to maintain the changelog and releases, but I don't mind doing it, honestly.

My personal opinion/experience is that stable (i.e., tagged) releases should be a bit less frequent than following master to give time for bugs to bubble up to the surface.  Some users like following the bleeding edge (master) and others will be put off if they experience bugs too frequently.

If you're OK with this model, I can update the README to point users by default at the tagged releases instead.  It also removes git from the equation by default, which is something less-experienced users have a lot of trouble with.

digiltd:
There are various changelog generators, [here is one based on this repo ](https://gist.github.com/digiltd/e5ffe8303218086b99b9b152a342f3f6) using [this generator](https://github.com/skywinder/Github-Changelog-Generator).

Usually not a fan of the "instant fix" but by adding labels to issues and requests etc gives a fair bit of control. [(page of projects using it)](https://github.com/skywinder/Github-Changelog-Generator/wiki/Projects-using-Github-Changelog-Generator) 

Regarding releases, another option would be to use branches to separate the development and stable releases.

jingai:
I do already tag releases (which can be checked out).

Typically master is the development branch.

m0ngr31:
Let's move the changelog discussion to [here](https://github.com/m0ngr31/kodi-alexa/issues/63)

PR #60, opened 2016-11-10
Make python-Levenshtein module opt-in

jagsta:
As discussed, mandating python-Levenshtein is going to cause more problems than it solves for most users. Section added to README to describe how to include it if desired, and removed from requirements.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,15 @@ You can also enable some extra verification of the SSL certificate by setting `S
 Note that you will need to manually install the dependencies in `requirements_verify.txt` if you want this additional validation.  Depending on your host environment, something like `pip install -r requirements_verify.txt` might work.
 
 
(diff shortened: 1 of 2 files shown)


jingai:
Looks good.

But it might be worth mentioning that AWS doesn't support it. The way it's worded suggests it's possible. Even though it may be through some unofficial way, if we suggest it we are likely to get asked about it. 

People tend to latch onto anything that speeds things up, afterall.

jagsta:
I'm not sure about Heroku supporting this either, perhaps I should say this is only supported when the skill is locally hosted as a WSGI script?

jingai:
I don't know either. You could just say as much in the README if you want. If someone tries it and it doesn't work, we can change it then.

jingai:
Just checked your update. What you said is sufficient I think.

PR #61, opened 2016-11-11
Support playing random movie by genre

jagsta:
From the TODO features list, this PR provides support for playing back a random movie by genre

I've modified the existing PlayRandomMovie intent to include an optional genre and the associated alexa_play_random_movie function in wsgi.py to limit selection of a movie to a specific genre when one is provided.

I've taken a steer from the ongoing discussion about CHANGELOG versions and left that as is.

I've got a branch of slot-generator which contains what I *think* are the changes required to generate the additional slot using the web app, although I can't easily test those. I'll submit these as a PR to @m0ngr31's slot-generator which will need to be taken with a pinch of salt.


jingai:
Looks good.  I'll test this and https://github.com/m0ngr31/slot-generator/pull/6 just as soon as I get some free time.

jingai:
Working great here.  Thanks for the contribution!

jagsta:
Thanks @jingai, glad to contribute. I've merged and squashed @digiltd's fixes into my branch of slot-generator, so this may now be OK (Not really knowing what i'm doing with slot-generator notwithstanding)

PR #62, opened 2016-11-11
I noticed the forum link wasn't working with https.

jacobhyman:
I noticed the support forum link wasn't working, so I thought it would be good to fix.  I am also interested in contributing in the future and just wanted to try out the basic workflow with something simple and non-functional.


m0ngr31:
I'm not sure why https isn't working anymore, so rolling with http is find with me.

PR #68, opened 2016-11-22
Solved update/clean libraries issue with some sleeps

m0ngr31:
Turns out to be an interesting issue:

- If an update library request is running, clean library will not run. Kodi doesn't queue it or anything
- If you run an clean library, it can take several minutes and time out the session with Alexa
- If you run the clean library method on a different thread (daemonized), you have to have some time for it to start running before the process ends or it won't fire in time

To alleviate these issues for now, I added some sleeps. If someone has a better way, just let me know.

### DIFF
--- a/wsgi.py
+++ b/wsgi.py
@@ -35,7 +35,8 @@ import random
 import re
 import string
 import sys
-import threading
+import time
+from multiprocessing import Process
 from yaep import populate_env
 from yaep import env
 
@@ -1007,13 +1008,14 @@ def alexa_clean_video(slots):
 
   kodi.UpdateVideo()
 
+  time.sleep(5)
+
   # Use threading to solve the call from returing too late
-  c = threading.Thread(target=kodi.CleanVideo)
+  c = Process(target=kodi.CleanVideo)
   c.daemon = True
   c.start()
 
-  # Calling this because for some reason it won't fire until the next command happens?
-  kodi.Home()
+  time.sleep(1)
 
   answer = "Cleaning video library"
   return build_alexa_response(answer, card_title)
@@ -1039,13 +1041,14 @@ def alexa_clean_audio(slots):
 
   kodi.UpdateMusic()
 
+  time.sleep(5)
+
   #Use threading to solve the call from returing too late
-  c = threading.Thread(target=kodi.CleanMusic)
+  c = Process(target=kodi.CleanMusic)
   c.daemon = True
   c.start()
 
-  #Calling this because for some reason it won't fire until the next command happens?
-  kodi.Home()
+  time.sleep(1)
 
   answer = "Cleaning audio library"
   return build_alexa_response(answer, card_title)


jingai:
Give me a couple days to get to this -- with the holidays here, I'm being pulled in other directions :)

m0ngr31:
I know that game. Just look at it when you get a chance.

jingai:
Seems to be working fine here.  Anyone else want to test before merge?

jingai:
Merging tonight.. just trying to give @edru a chance to try this, since he/she opened the original Issue.

40wattrange:
Will test when I get home tonight.

40wattrange:
Copying comment from the other post:

Just updated, seems the behavior is the same. Home screen, video library updates. No cleaning dialog appears and it jumps straight to updating.

Thought that I'd add that i dont think the library can be touched at the same time. I think if its cleaning it cant update, and if its updating it cant clean. You can test this by updating your library, and as its updating try to manually mark something as (un)watched, it wont let you. Dunno if that matters with what you are doing, but remembered you stated something about it being multi threaded - so maybe its trying to do both at the same time.

If this makes a difference I am running Kodi on the latest version of Android (a Shield TV)

m0ngr31:
> Just updated, seems the behavior is the same. Home screen, video library updates. No cleaning dialog appears and it jumps straight to updating.

Are you saying that it takes you to the home screen? If so, I removed that functionality, so maybe you're running on the old code still.

How big is your video library? The way it's setup now is to fire off the update library call, wait 5 seconds, and then fire off the clean library call. If it's not finishing updating your library in the 5 seconds I suppose that could cause an issue. I have a decent sized library, but updating it usually takes less than a second (but it's on a pretty beefy computer).

jingai:
If it's scraping in newly-added items, it can take a lot longer.  Plus if the machine is a client (media is not local) and you're using MySQL, it can take a while even without new items to scrape.

I'm not sure it's safe to assume it will complete within any set amount of time.

That said, it seems to work here for me even when the update does take longer than 5 seconds.  I'm guessing Kodi queues up the clean library call until after the update finishes.

40wattrange:
I'll see if I  an update again and see if the behavior changes. My library
is roughly 6TB large on a media server and a sql library backend, all on
ethernet, gigabit Wan if it needs to reach out for scraping info, and still
takes a good 45 seconds to a minute.

The original concept of this worked very well, it just cleaned :).

I'll move my instance over and resync from empty, then try again.

40wattrange:
- Stopped apache
- renamed kodi-alexa folder
- synced new copy from git master
- modified IP to connect properly to media center
- Started apache
- "Alexa, tell the media center to clean the video library" (saying just "library" defaults to audio now, where as it used to default to video
- No more home activity, but the updating of the library occurs immediately, and that's it. After the update, nothing further happens. Confirmed this with a new show in which a higher quality was downloaded. The older entry never removed itself after the command.
- Rebooted apache server for good measure
- Behavior still the same

So all in all "clean video library" just does a video library update, but the "home" action no longer occurs.

It was mentioned that this may be a threaded activity, not sure if you are trying to clean and update at the same time, but I know that it is not possible. KODI is a weirdo about allowing 2 things to happen w/ the library at once. (just try to do it manually and you will see it wont allow it).

Setup:
Kodi installed on NVIDIA Shield TV on Android 6.0
~6 TB of media on a virtual ubuntu server 16 (local) (OS on SSD, media all on spinning disk w/ 10 GB SSD cache)
Library is hosted on a virtual ubuntu server 16 w/ SQL (pure SSD)
All connections running over CAT5e (no wifi involved)
1 Gb WAN if it has to reach out to the web for anything

PR #73, opened 2016-11-30
Added timeout to Kodi POST call

onegambler:
* Added timeout parameter in the post call and added configuration in env file.
* Refactored environment values with default when missing


jingai:
Just curious, have you actually seen this happen?

m0ngr31:
I've seen timeouts on Heroku. There is a hard 30 second timeout there and it'd fail sometimes if the dyno was idling and my external drive needed to spin up. Haven't seen it with lambda yet though.

jingai:
Lambda has its own timeout as well, which would catch this.

I was mostly wondering if he/she had seen Kodi hang the connection indefinitely (as opposed to closing it or crashing, effectively closing it).  I'd be surprised if Kodi doesn't have its own send timeout, so I'd think to hit this case Kodi would have to actually hang?

kodi.py
```diff
@@ -109,24 +106,13 @@ def SendCommand(command):
...
-  PASS = os.getenv('KODI_PASSWORD')
-  if not PASS or PASS == 'None':
-    PASS = 'kodi'
+  SCHEME = os.getenv('KODI_SCHEME', 'http')
+  SUBPATH = os.getenv('KODI_SUBPATH', '')
+  KODI = os.getenv('KODI_ADDRESS', '127.0.0.1')
+  PORT = os.getenv('KODI_PORT', '8080')
+  USER = os.getenv('KODI_USERNAME', 'kodi')
+  PASS = os.getenv('KODI_PASSWORD', 'kodi')
+  KODI_RPC_TIMEOUT = os.getenv('KODI_RPC_TIMEOUT', '60')
```
jingai: All of this was done intentionally because in Lambda deployments, unset environment variables are set to the string 'None' rather than the None type, as the comment above the code block describes.

jingai:
It's not a bad idea to be safe and timeout here anyway, but I'm not sure it needs to be a configurable parameter.  The user is waiting for feedback as this is happening, so I'd say something like 30 seconds is probably the most anyone would sit through without assuming it just failed.

onegambler:
Hi,

I am using it with Lambda and I do see timeouts. As you were saying it doesn't hang, but I receive a generic
 "there's something wrong with your skill". I think it would be better to have a proper Alexa message saying "I can't connect to kodi". So timing out before the lambda and handling the error.

Apologies if the change on the getEnv are wrong. I will rollback it.
Please let me know if you want me to make the timeout non configurable and set it to 30 seconds.

jingai:
Doesn't Alexa herself time out after 10 seconds though?  So you won't get a better response out her with this.

Having a timeout when talking to Kodi just to prevent the server thread from hanging (if you're hosting the skill yourself somewhere that doesn't have a hard timeout) isn't a horrible idea, I just don't see how it'd ever trigger unless Kodi itself hangs after the connection is opened.  More specifically, the web server thread in Kodi would have to hang, and only *after* the connection has already been established.

If the entire application (Kodi) had hung, I'd be inclined to restart it anyway (which would close the connection with the skill) since it'd be unusable.

That's why I was asking if you'd specifically seen the web server thread in Kodi hang.

jingai:
I just tested it by adding a ```time.sleep(10)``` to the entry point of the skill, and she responded with:

```The remote endpoint could not be called, or the response it returned was invalid.```

Sleeping less than 10 seconds returns the response I expected.

m0ngr31:
> Doesn't Alexa herself time out after 10 seconds though? So you won't get a better response out her with this.

Alexa should only timeout if the connection to the skill dies. So whatever the timeout is configured on Lambda, 30 seconds on heroku, ect.

Can't say I've ever had an issue with Kodi timing out though.

jingai:
Hmm, I'm getting her error response if I sleep more than 10 seconds.  I'll play around more when I get home.

onegambler:
I've seen this happening on my Alexa. I have 60 seconds timeout on a Lambda and it does happen that just hangs there for a while and I get back a "something went wrong".

I am still investigating on the problem, not sure if it's related to slow responses from my Kodi or something else. I just thought it was a good idea to set a timeout when making the call.

I didn't know that Alexa was timing out after 10 seconds. Having a look at the metrics, I have executions longer than that.

jingai:
The skill will finish executing (or hit Lambda's 60 second timeout) no matter what.  I'm pretty sure Alexa has a separate timeout for voice responses, though.. but maybe it's different for Lambda.

Playing with things now.

onegambler:
@jingai you are right. Alexa has 10 seconds timeout.

https://forums.developer.amazon.com/questions/9176/what-is-the-timeout-of-the-request-sent-by-alexa.html

jingai:
Yeah, now that I'm home, I just tried it.  I added a 10 second sleep to the skill and got:

```The requested skill took too long to respond```

So, I don't think you can accomplish what you wanted unless you set the RPC timeout to < 10 seconds.  I wouldn't recommend that, though.

Furthermore, a lot of the intents issue more than one call to Kodi, so it's not quite as simple as just adding a timeout to each call.

onegambler:
I see what you mean. The exception is not propagated, in case of error will return a dictionary. So the execution will continue.

OK, I will close the pull request.

Thanks for this skill!

jingai:
Hey, thanks for the PR nevertheless!  It was good to at least have a discussion about it.

jingai:
Also, if you need help figuring out why you're seeing such absurdly long execution times (nothing should take 60 seconds), feel free to post on the forum and we'll see what we can figure out.

onegambler:
Thanks! 

I will try to fiddle around a little bit more and see if I can figure it out on my own first.

Cheers!

jingai:
For the sake of documenting this, I figured I'd add that in the card, she does specifically state that:

```Applications must respond within 10 seconds of the request being sent```

But she does wait until the skill finishes (or some other timeout is triggered) before blurting out that there was a problem.  Seems like a bug on Amazon's end, honestly.  There is probably a reason they did it this way.. or they were just lazy.. who knows.

m0ngr31:
That's good to know. Thanks for finding that.

PR #77, opened 2016-12-04
Add the ability to turn on the kodi host using wake-on-lan

ta264:
Obviously this will only work if the kodi-alexa app is running on a different host to kodi.  I've added some information to the `.env` templates but can add something to the README too if you'd like.

### DIFF
--- a/.env.lambda
+++ b/.env.lambda
@@ -14,6 +14,15 @@ KODI_PORT =
 KODI_USERNAME =
 KODI_PASSWORD =
 
+# If you want to enable WOL, set these variables.
+# If your skill endpoint is on the same subnet as the kodi host, only set the mac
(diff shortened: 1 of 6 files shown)


jingai:
Thanks for the submission.

But, should this not be a separate Smart Home skill?  From my own experience, Alexa is going to have trouble with this using those reserved keywords (like "turn on").

Plus, I honestly think this is a bit of scope-creep.  The skill here specifically communicates with Kodi.  and I'm not sure it's a good idea to start extending it to all the bits you can wrap the media center with (OS commands, lighting, power, etc).

jingai:
I believe you could also use HA Bridge to accomplish this.

ta264:
Thanks, you're probably right. I'll have a go at separating this out into a separate smart home skill.

PR #85, opened 2016-12-30
Initial commit for move to flask-ask.

m0ngr31:
This is a big update and the precursor to v2.5!

I have moved to flask-ask to help make the skill as clean as possible. This move allows the skill easier compatibility, and I believe it will make translation work very trivial since everything is in a template file now. Using something like babel should allow easy translation going forward (German first. We'll see what other languages Amazon will expand to soon).

I have moved the necessary code from `wsgi.py` to `alexa.py` and the main `application` method is now just `app`. So there will be some changes needed to upgrade current versions.

This is not ready for a merge into `master` yet due to documentation needing to be updated. However it is ready for testing.

Please let me know if you have any questions.


digiltd:
You have been busy :)

Could this also be a good time refactor the code into smaller modules?

The language issue is definitely something that we need to make sure is able to be addressed and if possible get it working for the initial release.

I will have a test and make some suggestions (I had a look at a flask-ask conversion a little while ago also).

How does the new .env file work in regards to timezones?

m0ngr31:
> How does the new .env file work in regards to timezones?

Timezones shouldn't be needed anymore since we're not verifiying the timestamp. According to this https://alexatutorial.com/flask-ask/configuration.html we shouldn't be at least. I guess if you want to enable it, it'd still figure it out without the timezone data.

> Could this also be a good time refactor the code into smaller modules?

I actually don't mind how it is now that the WSGI and Lambda still is ripped out of alexa.py. It's just intents in there now which is how it should be.

> The language issue is definitely something that we need to make sure is able to be addressed and if possible get it working for the initial release.

Yeah, it's been something I keep getting emails about and I'm excited to get some good stuff going for our German speaking friends.

jingai:
Time zone is used to determine the local end time of the currently playing item.

m0ngr31:
Okay, that's my bad. We can add that back in no problem then.

.env.example
```diff
@@ -14,8 +14,7 @@ KODI_PORT=
 KODI_USERNAME=
 KODI_PASSWORD=
 
-SKILL_APPID=
-SKILL_VERIFY_CERT=
```
jingai, line 18: I see that Flask-Ask enables cert verification by default, but do we want to provide a way to disable it?

.env.example
```diff
@@ -14,8 +14,7 @@ KODI_PORT=
...
+# SKILL_APPID=
```
jingai: Just so I understand, was there a reason for commenting this out rather than just leaving it empty in the example?

jingai:
Trying to check this out, but.. how do you call a Flask-Ask app from mod_wsgi?

The Flask documentation says to just import it as application:

```from alexa import app as application```

But that is giving me nothing at all.  I tried importing the Ask object, but it says that object is not callable.

**Comment by jingai** [comment]
Digging a bit further, it seems it's kind of working.. but I'm getting this odd routing exception:

```FormDataRoutingRedirect: A request was sent to this URL (https://mydomain/kodi-alexa/) but a redirect was issued automatically by the routing system to "https://mydomain/kodi-alexa/".  Make sure to directly send your POST-request to this URL since we can't make browsers or HTTP clients redirect with form data reliably or without user interaction.```

I don't really understand Flask well enough yet to know why, though.

**Comment by jingai** [comment]
Ok, that pisses me off and leaves a bad taste in my mouth, frankly.. apparently, it needed the final slash in the URL.... several hours later and it's something stupid, of course.. but still, their 'routing system' is clearly not very smart..

**Comment by jingai** [comment]
Went through every Intent systematically and things are mostly working.  Only encountered a failure on the WhatNewAlbums intent:

```[Sat Dec 31 20:34:02.575985 2016] [wsgi:error] [pid 15066] [client 72.21.217.78:29336] UnicodeEncodeError: 'ascii' codec can't encode character u'\\xef' in position 33: ordinal not in range(128)```

Haven't looked into it really, but it seems we might need to sanitize the responses some.

**Comment by m0ngr31** [comment]
> Ok, that pisses me off and leaves a bad taste in my mouth, frankly.. apparently, it needed the final slash in the URL.... several hours later and it's something stupid, of course.. but still, their 'routing system' is clearly not very smart..

Sorry to hear that. I'm hoping the switch will ultimately be worth the initial hassle though. The lambda process is a bit different too.

> Went through every Intent systematically and things are mostly working.

Thanks for doing that. I haven't made it through every single one myself.

> Haven't looked into it really, but it seems we might need to sanitize the responses some.

That doesn't surprise me that much. This is as good of a time to do it as ever.

**Comment by jingai** [comment]
Yeah, it's a positive change. The Flask documentation's just not the greatest to be honest and am a bit worried about debugging issues when they arise for others.

Mostly it'd be nice if it provided output for at least errors by default without having to enable debug mode.  I might have missed that in the documentation, though.

As for the responses, it seems something just got converted to ASCII along the way that should have remained unicode.  If you don't spot it within the week, I'll have a look.  Have to go back to the real job tomorrow though and am likely to be kept busy there for a little while.

File: speech_assets/SampleUtterances.txt [hunk 3]
```diff
@@ -334,6 +334,9 @@ NewShowInquiry if we have any episodes of {Show}
 NewShowInquiry if we have any new episodes of {Show}
 NewShowInquiry if we have any {Show}
 NewShowInquiry if we have {Show}
+OpenRemote open controller
+OpenRemote open controls
+OpenRemote open remote
```
jingai: Bear in mind that I had used "open" as the command to execute addons, so this means any addon with the name of controller, controls, and remote won't open via voice command.

edit: it appears "open menu" would have the same issue.

m0ngr31:
> As for the responses, it seems something just got converted to ASCII along the way that should have remained unicode.

I think I fixed this in my last commit.

jingai:
> > As for the responses, it seems something just got converted to ASCII along the way that should have remained unicode.
>
> I think I fixed this in my last commit.

On my list of things to check when I get off work.

jingai:
I think we may have hit a limit with regard to how big the Intent model can be.  I can't be sure, of course, but I think every time you reference a slot in the Intent model, it creates a copy of it and there's some overall limit that (at least to my knowledge) Amazon hasn't exposed.  Probably because they thought no one would hit it :)

If I try to save the current Intent model, it appears to succeed, though it takes a really long time to do so.  However, when using the simulator, I get:

```Error: No intent model has been built for this skill```

If I remove my songs from the MUSICSONGS slot (simply because it's the largest list, sitting at 26000 entries), everything works again.

Not really sure what to do about this, since users' libraries are lopsided in different ways; i.e., mine has many more songs, but others might have a crazy number of shows, or movies, etc.

File: music.py [hunk 4]
```diff
@@ -0,0 +1,84 @@
+import os
+import collections
+
+from pymongo import MongoClient
```
jingai: You'll probably have some users complain that pymongo isn't an optional dependency, just FYI.

jingai:
The WhatNewAlbums Intent still produces the same encoding error as before.  I suspect any other Intent with non-ASCII characters in the response will too.

```Newly added albums
Sending request to http://localhost:8080/jsonrpc
[2017-01-03 18:02:53,799] ERROR in app: Exception on / [POST]
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1517, in handle_user_exception  
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python2.7/dist-packages/flask_ask/core.py", line 516, in _flask_view_func
    result = self._map_intent_to_view_func(self.request.intent)()
  File "/var/www/ha/echo/kodi-alexa/alexa.py", line 1746, in alexa_what_new_albums
    return statement(response_text).simple_card(card_title, response_text)
  File "/usr/local/lib/python2.7/dist-packages/flask_ask/models.py", line 106, in __init__   
    super(statement, self).__init__(speech)
  File "/usr/local/lib/python2.7/dist-packages/flask_ask/models.py", line 53, in __init__
    'outputSpeech': _output_speech(speech)
  File "/usr/local/lib/python2.7/dist-packages/flask_ask/models.py", line 244, in _output_speech
    xmldoc = ElementTree.fromstring(speech)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1311, in XML
    parser.feed(text)
  File "/usr/lib/python2.7/xml/etree/ElementTree.py", line 1651, in feed
    self._parser.Parse(data, 0)
UnicodeEncodeError: 'ascii' codec can't encode character u'\\xed' in position 51: ordinal not in range(128)```

**Comment by m0ngr31** [comment]
> Not really sure what to do about this, since users' libraries are lopsided in different ways; i.e., mine has many more songs, but others might have a crazy number of shows, or movies, etc.

Might just need to limit the Slot Generator to 10k or something and see if that works. It's not ideal, but it'll give us a decent idea about how many they allow.

> The WhatNewAlbums Intent still produces the same encoding error as before. I suspect any other Intent with non-ASCII characters in the response will too.

Try it again after the commit I just made. I think it's a special case on this intent.

**Comment by m0ngr31** [comment]
Okay, I added German translations for card and voice responses (just used Google Translate, so I'm sure it's awful), but it's working and automatically translates if it's configured as a German skill.

Also, last night I pushed up some new intents that handle streaming party mode and all music by an artist to your Echo/Dot/Tap. This requires a bit of setup. You have to configure a mongoDB server and add information about it to the .env file as well as accept a warning I put in there about passwords being sent over the wire in plaintext.

One of the biggest hassles with steaming music is that Amazon won't allow non HTTPS urls, and if it has an SSL cert, it can't be self-signed. For a lot of people this is too complex, so I wrote a little HTTPS proxy that'll work (again, if they accept the warning message). It's configured so that you can run your own proxy (mine is just here on Github), or you can just have a Let's Encrypt cert on your local machine if you are running an reverse-proxy for Kodi already.

**Comment by jingai** [comment]
> > Not really sure what to do about this, since users' libraries are lopsided in different ways; i.e., mine has many more songs, but others might have a crazy number of shows, or movies, etc.
>
> Might just need to limit the Slot Generator to 10k or something and see if that works. It's not ideal, but it'll give us a decent idea about how many they allow.

The problem is that I don't believe it's a limit on any given slot, or even the total of all slots.  I think it's the cumulative number of entries in the slot *references* by the Intent model.  That is, I think simply referencing slots increases some counter (or maybe it's just an overflow, who knows) that has a ceiling on it that Amazon has not disclosed.

I think this because I tested it pretty thoroughly a little while ago while I was writing the generic search/play code.  The 50k items in slots limit is published by Amazon, but I don't have 50k items in my slots in total.  However, adding a new Intent that referenced existing slots caused it to fail, just as what you added makes it fail in the same manor on my installation.

Basically, I think the Intent model has some undisclosed maximum size, and Amazon isn't referencing the slots, but instead copying them into the model each time they're referenced.

**Comment by jingai** [comment]
> > The WhatNewAlbums Intent still produces the same encoding error as before. I suspect any other Intent with non-ASCII characters in the response will too.
>
> Try it again after the commit I just made. I think it's a special case on this intent.

It is indeed fixed, though it adds 'and' in between every item:

```You have Storm by Vanessa Mae, and  Ölürüm Sana by Tarkan, and  She Was a Boy by Yael Naïm, and  Peace And Love, Inc. by Information Society, and  Love In The Time Of Science by Emilíana Torrini, and more```

**Comment by m0ngr31** [comment]
I saw that. It's just a logic issue. Should be a pretty easy fix.

**Comment by m0ngr31** [comment]
Give it a shot now.

**Comment by jingai** [comment]
Yup, that's good now.

**Comment by jingai** [comment]
> Also, last night I pushed up some new intents that handle streaming party mode and all music by an artist to your Echo/Dot/Tap.

It'd be good to factor out the generic code for this so we don't duplicate all of the existing music playback methods.  But it'd probably be better to close out the Flask-Ask changes before doing so, because I've already got code ready for more generic playback that refactors some of it.

I was holding off on releasing the generic search/play code because of the Intent model size issue though.  I think before we can merge this branch we need to figure out how to properly solve that.

**Comment by jingai** [comment]
> Okay, I added German translations for card and voice responses (just used Google Translate, so I'm sure it's awful), but it's working and automatically translates if it's configured as a German skill.

@ausweider provided some initial effort for German here: https://github.com/ausweider/kodi-alexa/commit/03e131329ea85700baf7ecda45a9039a5da09d2d

Could have a look at how he/she chose to translate the strings since he/she is a native speaker.  It's a bit outdated, but could provide some direction.

File: .env.example (hunk 1 above)
Comment (m0ngr31) [reply .env.example:18]: If you don't provide a SKILL_APPID, I don't turn on verification.

File: music.py (hunk 4 above)
Comment (m0ngr31) [reply music.py]: Yeah, I'll work this into the if check to see if they've agreed to the warnings.

File: .env.example (hunk 1 above)
Comment (jingai) [reply .env.example:18]: They are two separate things though.  SKILL_APPID is to verify that the application ID matches what we expect, whereas SKILL_VERIFY_CERT is for verifying the certificate.

I personally want them both enabled at all times, but others might just want one or the other.

Comment (jingai) [reply .env.example:18]: I also don't see where you explicitly enable cert verification?  From the documentation (and I've verified it in execution too), Flask-Ask enables it by default.

Wasn't sure if we wanted a way to disable it, though.  Personally I'd never turn it off, so it's no big deal to me if you'd rather just let Flask-Ask handle it.

Comment (m0ngr31) [reply .env.example:18]: Well you know more than I do about the verification stuff. I haven't done anything with it. I would have to assume though that flask-ask takes care of the cert verification. Although I wasn't aware of it until you mentioned it.

I'm manually adding the APPID verification if the env variable exists though.

Comment (jingai) [reply .env.example:18]: Flask-Ask does deal with the cert verification -- in fact, I stole the code from there for the stuff I had in here initially ;)

I am not sure if it disables it for Lambda deployments automatically, but that could be one reason for leaving SKILL_VERIFY_CERT in place, so Lambda users can disable it (since it's redundant).

You'd have to check your own deployment though to see if it's verifying the certs.  I just put some print statements in the verifier code to see that it's executing.

Comment (m0ngr31) [reply .env.example:18]: Well the lambda deployment is totally different now. It is just a straight WSGI app now with now funny lambda tricks to get it working.

Comment (jingai) [reply .env.example:18]: Ahh.. but Amazon still verifies the cert before executing the Lambda function, right?

Comment (m0ngr31) [reply .env.example:18]: Nope, Amazon isn't doing anything with it now. Everything is routed through API Gateway and it's treated the same as if it wasn't a real lambda as far as I can tell.

Comment (m0ngr31) [reply .env.example:18]: Using this now: https://github.com/Miserlou/Zappa

Comment (jingai) [reply .env.example:18]: Oh ok.  So we probably want the cert verification enabled in Flask-Ask at all times anyway.

Comment (m0ngr31) [reply .env.example:18]: I can't think of any reason not to have it enabled now.

**Comment by digiltd** [comment]
Re Slots and dealing with 20k of them. 

The ASK documentation is vague, I might have the wrong end of the stick completely, I would be interested to discuss...

From what I have read it is my understanding that the purpose of a slot is to help identify and extract particular parts from the entire sentence. A number as a number, a name as a name, a song as a song etc. And to help identify what you are looking for you provide examples for what each slot might contain. Alexa uses the examples to "intelligently" match similar patterns.

It is then up to the Skill to deal with the contents of the heard slot. All Alexa does is identify the song was a song and not a town name.

Forgetting all the different complicated ways Alexa analyses the data she is trying to pattern match, instead lets assume that she is only looking at word count.

A simple example is you have a slot of 10 song titles that were all exactly five words long, and a slot of 10 artists that were all exactly two words long. When you tell the Skill to "play radio gaga", chances are it will identify "radio gaga" as an artist and send the "search artists" request to Kodi and end up playing Lady Gaga. 

If you swapped one of the five word song titles with a two word title (so 1x two-word and 9x five-word), then it would (I assume) still identify "radio gaga" as an artist. Alexa could see that there is a small percentage of two word songs but based on what she knows (100% of artists contain two words, 90% of song titles five words and only 10% of song titles are two words) she would have to treat it as an artist.

This is not the best example as the problem (as I am sure you know) with songs, artists and films as they are all super varied and you could easily swap a list of songs with a list of films and not spot the differences.

@jingai have you tried (can you try?) using a curated slot list of 200 songs/artists/albums that covered a broad spectrum of each and then tested it with examples that were not in the slot lists?

**Comment by jingai** [comment]
The problem is I wouldn't really know how to curate it without fully understanding how Amazon uses them.  Their descriptions of how it matches things to slots is vague at best.

I was thinking of modifying the slot generator to choose N (200, 1000, 5000, whatever works; bigger is better in my experience, so it'd be matter of choosing N where the skill doesn't break) random items for each slot.  But I need to set aside a decent chunk of time to test it.

If string length is also a factor, we could distribute the random choices so we get a roughly equal distribution of samples at varying lengths too.

**Comment by jingai** [comment]
The other thing to think about is if a user doesn't have enough of a given media type in their library for the Slot to turn "generic".

Might need the Slot generator to spit out some canned entries in that case.

**Comment by digiltd** [comment]
Hopefully the new built-in intents will help when they arrive.

When the UK didn't have IFTTT integration with our Echos, we used a workaround where we could send any phrase to the Maker Channel (webhooks) on IFTTT. I used to use a list of phrases for the sample utterances which would allow me to pass any phrase to the channel.

hello
hi there
it's cold here
my name is Alexa
the weather outside is frightful
the meaning of life is forty-two
what is the weather in Miami Beach
you get nothing you lose good day sir
the quick brown fox jumps over the lazy dog
we are the music makers we are the dreamers of dreams

**Comment by jingai** [comment]
Yes, but we need to do something in the meantime so those with large libraries aren't faced with an odd failure case like this (Amazon telling the user the Intent model was built just fine, then failing at runtime).

Simply choosing, say, 2500 random items from the library for each slot would probably work for now.  Just need to test it.

**Comment by m0ngr31** [comment]
I could whip something like that up pretty quickly

**Comment by jingai** [comment]
Go for it.  It'd be a few days before I could get to it, and I think it'd be a release-stopper for this branch unless you felt like dealing with the storm on the forum lol

File: kodi.py [hunk 5]
```diff
@@ -91,11 +91,14 @@
...
   normalized = unicodedata.normalize('NFKD', media_name).encode('ASCII', 'ignore')
-  # Strip symbols
-  m = RE_NAME_WITH_PARAM.match(normalized)
-  if m:
-    return m.group(1)
-  return media_name
+
+  # Strip things between and including brackets and parenthesis
+  removed_paren = re.sub(r'\([^)]*\)', '', normalized)
+  removed_bracket = re.sub(r'\[[^)]*\]', '', removed_paren)
```
jingai: Should it actually strip out what's in the parentheses/brackets?  I ran into a problem with my playlists because of this before -- I had named them like:

Kids (Music Videos)
Kids (Shows)

I don't think I've seen a problem with other titles, but I can imagine it might be.

Particularly for playlists, though, it might cause failed matches merely because the user isn't likely to have enough in the slot to force it over to a "generic slot".

m0ngr31: Initially I put this in at the beginning because I was having problems matching things like 'The Office' because Kodi has it as 'The Office (US)' and there were some other shows like Archer that Kodi labels as 'Archer (2009)'. I guess we could just add an option for the function that will just strip out the actual characters or the whole thing depending on how you are running it.

jingai: Might try with those again -- I think the fuzzy matching will catch those, at least "The Office".  Archer maybe not, but I'd be curious to see if it does.

An option might be a good idea, if we find it's necessary.  Just off the top of my head, I can think of another instance where you may want to keep what's inside the parentheses:

Some Song (feat One Band)
Some Song (feat Another Band)

There are probably other cases where the user would want to speak what's inside the parentheses.

jingai: Plus, scrapers usually add those suffixes to disambiguate one title from another.  What would a user do if they had both "The Office (US)" and "The Office (UK)"?  Or "Archer (2009)" and "Archer (1975)"?

jingai:
Don't think there's any outstanding issues at this point.. I'll run through all of the Intents again tomorrow night and report any issues.

jingai:
Excellent job on the move to Flask-Ask btw.  It'll likely bring a whole new set of issues to deal with on the forums, but it was sorely needed and will be worth it in the long run.

I'll refactor my generic search/play stuff as soon as this is all merged into master.

jingai:
When we go to actually release 2.5, it might be prudent to put a note in the README and a post on the forum that users should re-generate their slots since I'm sure I'm not the only one with a ton of songs in their library.

jingai:
Is there anything in particular I should know if I wanted to set up MongoDB on my own server to test out the streaming stuff?  I'll try it out tomorrow night too if I get a chance.

m0ngr31:
> Is there anything in particular I should know if I wanted to set up MongoDB on my own server to test out the streaming stuff? I'll try it out tomorrow night too if I get a chance.

I didn't run it off of my own server, but on mLab, you have to have a username/pass for the database itself, and not just the overall server credentials. Also, the MONGODB_URL variable is just the domain, no HTTP or anything. If that makes sense.

> When we go to actually release 2.5, it might be prudent to put a note in the README and a post on the forum that users should re-generate their slots since I'm sure I'm not the only one with a ton of songs in their library.

Good call

> Don't think there's any outstanding issues at this point.. I'll run through all of the Intents again tomorrow night and report any issues.

I appreciate it.

> Excellent job on the move to Flask-Ask btw. It'll likely bring a whole new set of issues to deal with on the forums, but it was sorely needed and will be worth it in the long run.

I sure hope so. It's been on my mind to do for at least 6 months.

jingai:
We honestly should have a general section in the README too for upgrading.  I'll make an Issue for it.

danielmasur:
Awesome. I'll (He, as female would be Ausweiderin) hopefully get a chance to look at this over the weekend and do some german refining :)

- Besides: It should be noted that for AWS deployment to work one needs to change s3_bucket in  zappa_settings.json to something unique otherwise deployment fails with an error:
An error  occurred (Bucket already exists)

- Also AWS specific: readme suggest setting up a python virtual environment called venv. venv/ should therefore be in .gitignore by default

Issues I encountered during testing:
So I tell Alexa to ask kodi to play music by Metallica
"frage kodi spiele musik von metallica"

`{
  "session": {
    "sessionId": "SessionId.blabla",
    "application": {
      "applicationId": "amzn1.ask.skill.blabla"
    },
    "attributes": {},
    "user": {
      "userId": "amzn1.ask.account.blabla"
    },
    "new": true
  },
  "request": {
    "type": "IntentRequest",
    "requestId": "EdwRequestId.blabla",
    "locale": "de-DE",
    "timestamp": "2017-01-06T14:41:03Z",
    "intent": {
      "name": "ListenToArtist",
      "slots": {
        "Artist": {
          "name": "Artist",
          "value": "metallica"
        }
      }
    }
  },
  "version": "1.0"
}`

I get the Service Response:
The remote endpoint could not be called, or the response it returned was invalid.

But when I tell Alexa to ask kodi what is currently playing
"frage kodi was das ist"

`{
  "version": "1.0",
  "response": {
    "outputSpeech": {
      "type": "PlainText",
      "text": "The current song is Blitzkrieg, by Metallica, on the album Garage Inc.."
    },
    "card": {
      "content": "The current song is Blitzkrieg, by Metallica, on the album Garage Inc..",
      "title": "Momentan läuft",
      "type": "Simple"
    },
    "shouldEndSession": true
  },
  "sessionAttributes": {}
}`

This issues comes up intermittently.

m0ngr31:
> I get the Service Response:
The remote endpoint could not be called, or the response it returned was invalid.

As a test, could you try running your server from something like [ngrok](https://ngrok.com/) (You'll have to update the endpoint in your skills settings as well). I'd be curious to see the exact error.

You can run the server locally by doing something like this: `gunicorn alexa:app -b 0.0.0.0:8001` and then in another console window: `ngrok http 8001`.

jingai:
Sorry I'm a little late with this, but I checked every Intent again and all seems well.  Only found a minor typo on volume_set in templates.yaml, which I've already fixed and pushed up.

m0ngr31:
That's great. Thanks for doing that. Did you by chance try out the streaming?

jingai:
Oh, sorry, I forgot to mention that.  I have not yet set up streaming.

I might be able to squeeze in some time to set it all up on Thursday or Friday.

digiltd:
Regardless of #90 , it is becoming clear there are too many different intents using similar phrases for "playback control". There are also custom intents and built-in intents that will cause problems.

```
"intent": "AMAZON.StopIntent"
"intent": "AMAZON.NextIntent"
"intent": "AMAZON.PauseIntent"
"intent": "AMAZON.PreviousIntent"
"intent": "AMAZON.ResumeIntent"
"intent": "AMAZON.StartOverIntent"

"intent": "PlayPause"
"intent": "Prev"
"intent": "Skip"
"intent": "StartOver"
"intent": "SubtitlesNext"
"intent": "SubtitlesPrevious"
"intent": "AudioStreamNext"
"intent": "AudioStreamPrevious"
```

From a discussion about the removal of the single word utterance example "previous" 

> 
> "Alexa, ask kodi to next"
> "Alexa, tell kodi previous"
> 
> Testing all the prev and skip intents does highlight some clashes with the audio stream controls.
> 
> "to previous" = AudioStreamPrevious
> "to next" = Skip
> "previous" = AMAZON.PreviousIntent
> "next" = AMAZON.NextIntent
> "next track" = Skip
>

m0ngr31:
Does it actually clash with AMAZON.PauseIntent, AMAZON.ResumeIntent, AMAZON.StartOverIntent, AMAZON.NextIntent, and AMAZON.PreviousIntent specifically? I know some of the utterances overlap, but these intents specifically are used only when the stream is playing (or paused), and don't use the Kodi skill invocation.

digiltd:
Yes, it does when I test it. Does it not when you test using the same phrases I tested with above?

speech_assets/SampleUtterances.txt
```diff
@@ -1,3 +1,5 @@
+AMAZON.HelpIntent help
+AMAZON.HelpIntent open help
```
digiltd: You might not know, but you shouldn't need to add samples for the built in intents unless you wanted to extend the built in list of sample phrases with something specific.

It does make things easier as you don't have to provide translations for current and future language support. These standard built-in intents are already supported the three 

https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/built-in-intent-ref/standard-intents#available-standard-built-in-intents

jingai:
@m0ngr31, is there any way to set up streaming so it's not passing the username and password on the URL?  That's not something I really want to do.

edit: Nevermind -- I see that the request is coming directly from the Echo, so I can make this work :)

m0ngr31:
I couldn't think of a way to not include the user/pass combo, so let me know if you found a way.

speech_assets/SampleUtterances.txt
m0ngr31: I didn't know that. Thanks. I'll get that reverted.

jingai:
Been trying to make this (music streaming) work for a while now.  I have it to the point where the Echo requests the stream, but for some reason it can't actually get at the file via the VFS path.  Kodi sees the request, but it's not actually sending over the file.

In Kodi 17b6 on my Linux box, the URL it's trying to request returns "File not found" from Kodi.  On one of my Macs with Kodi 17rc3, the response body is empty.

I've tried accessing the URL via curl on both boxes too and get the same results.  I've done this both through the proxy I set up and straight to Kodi.

I'll play with it some more tomorrow, but if you have any ideas, I'd appreciate it.

jingai:
@m0ngr31, are you able to request a file via your web browser from Kodi?

Like:

http://user:pass@kodihost:8080/vfs/songfilepath

Does it return the file?

m0ngr31:
Yeah, it always returns a file.

m0ngr31:
The only thing I can think of right off the bat is that maybe your password has some symbols that need to be escaped or something?

jingai:
I can't figure out what I am missing then.. none of my Kodi installations return a file to me.  17b6 gives me an empty page with "File not found" in the body.. both of my 17rc3 clients give me empty bodies.

This is the entire response from Kodi 17b6 when I request the URL via curl:

```<html><head><title>File not found</title></head><body>File not found</body></html>```

I've url encoded the path, too.

> The only thing I can think of right off the bat is that maybe your password has some symbols that need to be escaped or something?

My password is long, but it is alphanumeric [A-z][0-9].

**Comment by m0ngr31** [comment]
How is your media stored? Locally or on a different computer?

**Comment by jingai** [comment]
Ahh, it appears the first slash in the path was stripped out somewhere.

I'm guessing maybe you use URIs in your sources like smb:// or whatever.  Mine are just paths on the disk.

So, for me, it was trying to play:

```/vfs/Volumes%2Fdata%2FMusic%2FAudio%2FAdult%2FTool%2FUndertow%2F03%20-%20Sober.m4a```

which is a relative path, and doesn't work.

But this does:

```/vfs/%2FVolumes%2Fdata%2FMusic%2FAudio%2FAdult%2FTool%2FUndertow%2F03%20-%20Sober.m4a```
```curl -u kodi:password "http://rakka:8080/vfs/%2FVolumes%2Fdata%2FMusic%2FAudio%2FAdult%2FTool%2FUndertow%2F03%20-%20Sober.m4a"```

It's also possible that mod_wsgi is stripping the second slash.. IIRC, it does do that.  So I'll probably have to work around it in the proxy.

**Comment by jingai** [comment]
Ok, I'm munging the URL in the proxy now and it's sending the correct URL to Kodi.  Now just to make the proxy actually work... Node is a PITA to get working with Apache, so I'm trying to get one going via mod_wsgi/Flask.. if you happen to have a link to one, let me know please :)

**Comment by m0ngr31** [comment]
I just updated the virtual path function if you want to try it again.

> Node is a PITA to get working with Apache, so I'm trying to get one going via mod_wsgi/Flask.. if you happen to have a link to one, let me know please :)

Unfortunately I've only used Node with nginx :\

Although, if you want one for flask, look at this: http://flask.pocoo.org/snippets/118/

File: kodi.py [hunk 7]
```diff
@@ -880,16 +880,15 @@ def SystemEjectMedia():
...
-  url = "%s://%s:%s@%s:%s/%s" % (SCHEME, USER, PASS, KODI, PORT, SUBPATH)
-
-  url = url + path
+  url = "%s://%s:%s@%s:%s/%s/vfs" % (SCHEME, USER, PASS, KODI, PORT, SUBPATH)
 
   # Remove any double slashes in the url
   url = http_normalize_slashes(url)
 
+  url = url + '/' + path
+
```
jingai: OH LOL.. I can't believe I missed this.. I was so focused on mod_wsgi being the culprit...

I'll test in an hour or so.

jingai:
> I just updated the virtual path function if you want to try it again.

This results in a valid path, and I can see the Echo requesting the URL.

> Although, if you want one for flask, look at this: http://flask.pocoo.org/snippets/118/

I saw that, but for some reason it isn't working.  Kodi reports that it sees the request, but it loses the connection:

```
14:20:53.879 T:140581728679680   DEBUG: CWebServer[8080]: request received for /vfs/%2FVolumes%2Fdata%2FMusic%2FAudio%2F
Adult%2FTool%2FLateralus%2F08%20-%20Ticks%20%26%20Leeches.m4a
14:20:53.917 T:140581728679680   DEBUG: CWebServer [MHD]: Failed to send data: Connection reset by peer
```

I'll get it figured out.. was just hoping I could find something pre-made so I didn't have to write it myself :P

danielmasur:
Another language based issue comes up. Sentence construction is vastly different from one language to the other. It's really important to be able to change the order of {VAR} either in a language based .yaml or later on in the .po. I suppose it works for now, but should be adressed at some point.

digiltd:
> Another language based issue comes up. Sentence construction is vastly different from one language to the other. It's really important to be able to change the order of {VAR} either in a language based .yaml or later on in the .po. I suppose it works for now, but should be adressed at some point.

@ausweider Hi, just a thought, is there any value in just translating the template file per language rather than translating at init using Babel?

The automatic translation is nice, but considering each user will have to set up the skill themselves, they are able to upload the relevant templates.yaml file depending on their language.

For my local install I have removed all the Babel bits and only use the translated templates.yaml file. 

The Babel package is 20mb, I don't know if performance takes a hit when it imports in alexa.py. The function size certainly does. I have also removed pycountry as that was another 10mb package. My lambda function has gone from 26mb to 8.7mb.

jingai:
Finally found my problem with music streaming.. the Content-Type must be set, and apparently Kodi doesn't do that.  At least, I can't see it in the request headers.  So, I'll need to find a way around this..

jingai:
Got it, finally :)

I had to set 'Content-Type' to 'audio' in the headers.  "Alexa, ask Kodi to stream music by artist" is now working.

At some point we should factor out the code in all of the Listen* intents though so we're not duplicating everything.  But it's a good first step!

m0ngr31:
Where'd you have to add content headers, on the flask proxy?

jingai:
Yeah, in the flask proxy.  I made the request from there to Kodi, but the headers that came back didn't include the content-type field.  I added it in before passing it all back to the Echo and it works now.

Without the content-type set, the Echo would just close the connection.

m0ngr31:
Interesting

jingai:
All of my music is .m4a, which may have something to do with it.

README.md
```diff
@@ -68,8 +84,6 @@ Once you have my repo cloned and you are in the directory, you can setup the fol
 
 You can do this easily from the command line: `heroku config:set KODI_ADDRESS='your_ip_or_dynamic_address' KODI_PORT='kodi_port' KODI_USERNAME='kodi_username' KODI_PASSWORD='kodi_password' --app app-name-and-number`. Changing of course for your settings. You can also use the settings page on your Heroku app to add these.
```
jingai: Should this example be updated to reflect all of the newer environment variables we have?

Or maybe a comment to continue reading the section that talks about the other variables at least, for those with short attention spans.

README.md
```diff
@@ -68,8 +84,6 @@ Once you have my repo cloned and you are in the directory, you can setup the fol
...
 
-You can also alternatively store configuration in a file called `.env` which you need to create yourself from a copy of `.env.wsgi`
-
```
jingai: Was this removed because it doesn't work?  If it does work, it seems like the easiest thing to document since we wouldn't have to update the example command-line above every time we change the variables.

README.md
```diff
@@ -81,85 +95,61 @@ Heroku doesn't just fire up the server automatically, so you have to tell it to:
...
+After you've done that, run `pip install virtualenv`. This is required for a later step.
 
->  _Lambda functions for Alexa skills must be hosted in the US East (N. Virginia) region. Currently, this is the only Lambda region the Alexa Skills Kit supports_.
+Now, clone my repo: `git clone https://github.com/m0ngr31/kodi-alexa.git` and `cd kodi-alexa`. Once you are inside the project directory, you're going to create a new "Virtual environement" and then activate it:
+`virtualenv venv` and `source venv/bin/activate` (if you are on Windows, that's `venv\Scripts\activate.bat` or `venv\Scripts\activate.ps1` for Powershell).
 
-The above statement is repeated many times throughout the docs.
+After successfull completion, run `pip install -r requirements.txt` and `pip install zappa`. Before you deploy, you need to copy the `.env.example` file to `.env` and enter the correct information for: KODI_ADDRESS, KODI_PORT, KODI_USERNAME, and KODI_PASSWORD. I'll go over the other variables in another section below.
```
jingai: It might be a good idea to avoid calling out specific variables here and simply put the documentation in the .env.example file.

jingai:
@m0ngr31, the next thing I have on my list to work on is extending the streaming support to work with all of the same variations that the ListenTo* intents currently do.

Just so we don't overlap work though, could you please let me know if you decide to start work on this?

README.md
m0ngr31: Heroku deploys based on what is in git. So since .env is in .gitignore, it won't be sent to Heroku.

jingai: Understood.

jingai:
Sometimes when trying to resume an audio stream I get the following:

```
[2017-01-15 17:17:43,664] ERROR in app: Exception on / [POST]
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1982, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1614, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1517, in handle_user_exception 
    reraise(exc_type, exc_value, tb)
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1612, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python2.7/dist-packages/flask/app.py", line 1598, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python2.7/dist-packages/flask_ask/core.py", line 544, in _flask_view_func
    self._update_stream()
  File "/usr/local/lib/python2.7/dist-packages/flask_ask/core.py", line 513, in _update_stream
    fresh_stream.__dict__.update(self._from_context())
TypeError: 'NoneType' object is not iterable
```

It looks like an issue with Flask-Ask, but I figured I'd let you know anyway.

If I keep repeating the command, it will eventually work.

digiltd (commented):
oops

speech_assets/SampleUtterances.txt
```diff
@@ -324,16 +324,12 @@ Mute toggle unmute
 Mute unmute
 NewShowInquiry if there are any episodes of {Show}
```
digiltd, line 383: That's the "ifs" sorted :)

but we should also add the various "is" and "do" examples

jingai:
> That's the "ifs" sorted :)
>
> but we should also add the various "is" and "do" examples

It'll happen.  Just trying to establish a baseline and push up as few changes as possible to get this branch mergeable.

speech_assets/SampleUtterances.txt
jingai: It'll happen. Just trying to establish a baseline and push up as few changes as possible to get this branch mergeable.

m0ngr31:
> Without the content-type set, the Echo would just close the connection.

Can you share the code for your proxy?

> The automatic translation is nice, but considering each user will have to set up the skill themselves, they are able to upload the relevant templates.yaml file depending on their language.

That's a good point. It might be easier to maintain going forward instead of having weird translation issues like we are going to be running into shortly from having variable names in strange places.

>  I have also removed pycountry as that was another 10mb package. My lambda function has gone from 26mb to 8.7mb.

If we have separate templates.yaml files, we can definitely rip out pybabel. I wonder how much effort it'd be to rip out pycounty too. It just looks like it's being used on 2 intents that probably aren't used much.

jingai:
As for pycountry, I do use it, so I'd prefer it if we retained the functionality.

pycountry itself is literally just a few lines of python.  On Debian, the pycountry package is 15KB (including README, changelogs, etc).  The language database isn't huge either, but it comes as part of the iso-codes package, which includes a lot of stuff we don't use.

I suppose you could just take pycountry's code as well as the ISO 3166-1 alpha-3 codes and include it in the skill.

The sizes we're talking about here are all pretty small these days.. I'm not sure you'll see any noticeable performance gains just by ripping out 17MB.  I'd personally rather import it and let others maintain it, though it probably isn't likely to change much over time.

edit: This is what I'm using for the proxy (very, very quick and dirty):

```
#!/usr/bin/env python

from flask import Flask
from flask import Response
from flask import request
from flask import stream_with_context

import requests
import urllib
import logging

app = Flask(__name__)
logging.basicConfig(level=logging.INFO)
CHUNK_SIZE = 1024
LOG = logging.getLogger("kproxy.py")

@app.route('/<path:url>')
def home(url):
  kodi_music_hack = False

  split_url = url.lstrip('http:/').split('/', 2)
  proxy_url = 'http://' + split_url[0]
  if len(split_url) > 1:
    proxy_url += '/' + split_url[1]
  if len(split_url) > 2 and 'vfs' in split_url[1] and '://' not in split_url[2] and split_url[2][0] != '/':
    # if it's a local path, force absolute
    split_url[2] = '/' + split_url[2]
  if len(split_url) > 2 and 'vfs' in split_url[1]:
    proxy_url += '/' + split_url[2]
    kodi_music_hack = True

  LOG.info("Proxy for: %s", proxy_url)

  req = requests.get(proxy_url, stream = True)
  headers = dict(req.headers)
  if not req.headers.get('content-type') and kodi_music_hack:
    headers['Content-Type'] = 'audio'
  LOG.info("Got %s response from %s", req.status_code, proxy_url)

  def generate():
    for chunk in req.iter_content(CHUNK_SIZE):
      yield chunk

  return Response(stream_with_context(generate()), headers = headers)
```

digiltd:
re pycountry, I am not sure if the numbers are the same, but I did test the difference with and without (though it was many many commits ago). "With pycountry" was adding around a second onto the response time on AWS Lambda.

From what I can tell the current use is to convert the three character country code into the full language name i.e. eng to English, fre to French, dan to Danish. This is so when you switch subtitles Alexa can respond with the more comprehensible  "English" rather than "eng".


I didn't want to PR just yet as it can perhaps wait until the new Flask version is up.  But I have put together a lightweight alternative if you want a look to see if it is suitable: https://github.com/digiltd/kodi-alexa/tree/flask-ask-remove-pycountry

jingai:
If it does indeed slow things down for Lambda/etc users to use pycountry and your replacement is functionally the same, it's fine with me.

m0ngr31:
I'm fine moving away from pycountry and babel. I think having a dedicated templates file for other languages will be simpler in the long run. I have just a couple of issues with your branch @digiltd, I'll try to make a PR for it in the next couple of days.

digiltd:
There is a small issue regarding using the current .env method for environment variables when deploying with Zappa that might want to be addressed at some point. The solution would probably overcomplicate setup for users to a point where the benefit is just not worth it (especially given the current level of users having trouble understanding the concept of a .env file).

I am aware that the sky is not falling, but I would feel happier knowing that others were at least aware of the issue.  

It is about environment variables and Zappa. The .env approach is not ideal with this method of deployment as the .env is uploaded as a regular plain text file that gets referenced by the populate_env functions. This wasn't a problem <2.5 as env variables were set at deployment and the .env file left behind.

For Zappa the recommended way to use remote environment variables is by putting a json file in a remote S3 Bucket and then the app pulls the variables in from that using IAM.

https://github.com/Miserlou/Zappa#remote-environment-variables

For my local dev version I am using the S3 Bucket method, and if it is of interest, here is a branch showing what that involves https://github.com/digiltd/kodi-alexa/tree/flask-ask-zappa-remote-env-vars

tbh I really don't think having the .env file in the function folder is the worst thing in the world. I also am not sure people would really care. Especially when users are excitedly opening up http ports to their local network without fully knowing what they are doing :) 

I also don't know how this affects local or Heroku installs. It certainly would be a pain having to have (and support) different deployment methods for each.

m0ngr31:
> I also don't know how this affects local or Heroku installs. It certainly would be a pain having to have (and support) different deployment methods for each.

Have to use a different method on Heroku already. Zappa already allows you to enter the env variables in the the zappa file as well, which would be easier than trying to setup an S3 account and document that.

digiltd:
any updates on 2.5?

m0ngr31:
There is a little I still need to do with music streaming and translation. But hopefully I can finish it up soon. I haven't had any free time lately.

jingai:
The copying of templates.XX.yaml to templates.yaml is causing me some grief.  My web server's user doesn't have write access to the directory (intentionally), so the bit of code that tries to copy it over unconditionally just fails here.

Maybe it would be better to do this in the deployment script instead?

m0ngr31:
I just noticed that I'm running into the same issue with Lambda as well. What do you think the best option would be?

jingai:
My first thought would be to do it in a prep/deployment script, a la deploy-to-lambda.py.  Or just add to the README that the user has to copy the file before deployment.

Could also have a default templates.yaml there for English?

jingai:
Hmm.. yeah, honestly I'd:

templates.en.yaml -> templates.yaml

and just add documentation for users of other locales to copy the correct one in.

m0ngr31:
I thought about that as well, but was trying to figure out how it'd work when they need to do a git pull. Maybe I'll work on a deployment and upgrade script.

jingai:
Could just document it as part of the upgrade procedure.

Either way, it's an extra step for users.  It could at least be one less step for English users.

m0ngr31:
I made a PR over on the Flask-Ask project (https://github.com/johnwheeler/flask-ask/pull/112) so that we can keep the separate files & not have to worry about moving stuff around on a git pull when someone wants to update.

jingai:
Oh, good on you for going upstream :)

jingai:
Wait.. did we lose the German translations for responses?

m0ngr31:
Yes, they will have to be done again. But it'll be better in the long run so that the dynamic parts of the sentence will need to be in a different order depending on the language.

jingai:
We shouldn't pull over what we already had?

If not.. @ausweider, if you're still around.. help please :)

m0ngr31:
It's still in the git history if he need it, but I'm guessing it's going to be enough different that it won't help much.

jingai:
Oh, I thought it was relatively up-to-date before this point.  I gotcha.

jingai:
Posted on the Kodi forum asking for a translator.

README.md
jingai: Does this need to be updated?

m0ngr31: I actually have a pretty nifty solution for Heroku deploys coming really soon.

jingai: Ok.  Was just trying to ensure we didn't miss anything :)

jingai:
Commenting here to bump the flaskask version in requirements.txt to whatever version will include the custom template names.

jingai:
@sveni-lee and @ausweider, some new strings were added for music streaming intents, and they're currently in English only.  If you want to see them in German for the 2.5 release, please translate and submit a PR :)

jingai:
Rebased onto current master in preparation for merge.  If you have a local copy of this branch, you'll need to delete and re-pull.

jingai:
All that's left is testing now -- touched a lot recently by stripping out streaming music.

danielmasur:
Unfortunately I won't be available till after April 24th. @sveni-lee

PR #95, opened 2017-01-06
remove docker files

digiltd:
I have tried to get it to work and failed. I did some searching and couldn't find any instances of a successful installation using Docker. All I found were discussions about being unable to get it working.

https://lime-technology.com/forum/index.php?topic=53050.0
https://forum.libreelec.tv/thread-2135.html
https://forum.libreelec.tv/thread-1787.html


devdems:
I have a plan to get this working in docker so i will post back how this is done.

m0ngr31:
@snoopy86: That would be great thanks.

Procfile
```diff
@@ -1 +0,0 @@
-web: gunicorn alexa:app
```
m0ngr31, line 1: This file is needed for Heroku.

digiltd:
In the meantime do you want to keep the files that don't seem to be working available?

devdems:
It doesn't make a difference for me. When the final working version is ready i will upload it.

m0ngr31:
Okay @snoopy86, just make a new PR based off of the `flask-ask` branch when you get it going.

PR #96, opened 2017-01-06
reviewed german translations, added up to date german utterances, add…

danielmasur:
…ed venv/ to .gitignore


.gitignore
```diff
@@ -27,6 +27,7 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
+venv/
```
m0ngr31, line 30: Thanks for doing this. I use virtualenv-wrapper, so my venv folder in not in the project folder. I just didn't want to have to explain another step.

m0ngr31:
Thanks for doing this!

PR #120, opened 2017-03-05
[flask-ask] Fix BigForward/Backward and StartOver

zanix:
- `alexa_player_seek_bigforward` is defined twice, the second one should
be `alexa_player_seek_bigbackward`
- Change `PlayStartOver` use of `Player.GoTo` to `Player.Seek`
  - `Player.GoTo` doesn't work for non-playlist entries

Backport to master in PR #121


jingai:
cherry-picked this into flask-ask.

PR #121, opened 2017-03-05
[backport] Fix BigForward/Backward and StartOver

zanix:
- Backport from PR #120 
- `alexa_player_seek_bigforward` is defined twice, the second one should
be `alexa_player_seek_bigbackward`
- Change `PlayStartOver` use of `Player.GoTo` to `Player.Seek`
  - `Player.GoTo` doesn't work for non-playlist entries


jingai:
Thanks zanix.

PR #125, opened 2017-03-27
Remove pycountry

digiltd:
The large package size is causing the "Signature expired" error when deploying on a machine with a slow upload speed. As a quick fix I removed the pycountry requirement but I haven't added the lightweight fix used in my [flask-ask-remove-pycountry](https://github.com/digiltd/kodi-alexa/tree/flask-ask-remove-pycountry) branch. So for now it just reads the country code, not the country.

### DIFF
--- a/kodi.py
+++ b/kodi.py
@@ -36,7 +36,6 @@ import random
 import re
 import string
 import sys
-import pycountry
 from fuzzywuzzy import fuzz
 from fuzzywuzzy import process
 
@@ -943,7 +942,7 @@ def GetCurrentSubtitles():
   if curprops is not None:
     try:
       lang = curprops['currentsubtitle']['language']
-      subs = pycountry.languages.get(bibliographic=lang).name
+      subs = lang
       name = curprops['currentsubtitle']['name']
       if name:
         subs += " " + name
@@ -959,7 +958,7 @@ def GetCurrentAudioStream():
   if curprops is not None:
     try:
       lang = curprops['currentaudiostream']['language']
-      stream = pycountry.languages.get(bibliographic=lang).name
+      stream = lang
       name = curprops['currentaudiostream']['name']
       if name:
         stream += " " + name
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 requests
 gunicorn
 yaep
-pycountry
 pytz
 fuzzywuzzy


jingai:
Why not just go ahead and merge your branch so we don't just lose this functionality?  Is there something wrong with it?

digiltd:
the other branch is for the flask version and I haven't tested it with the old Skill

jingai:
I'll backport them I guess, otherwise the branches will diverge and there'll be a conflict when we go to merge back into master.

This is exactly the reason why I advocate master as a development branch.. I knew we'd be supporting two branches eventually.

Anyway, I'll make the changes on master and then @m0ngr31 can rebase the flask-ask branch to pull them into there.

jingai:
Added the lightweight alternative and pushed it up.  @m0ngr31 will need to rebase flask-ask onto master to pull the changes into there.

m0ngr31:
The only thing I didn't like about your branch on the flask-ask PR is using `# coding=utf-8`

It's just a personal preference thing, but if we could avoid using that, I'd prefer that.

digiltd:
Thanks for integrating the lighter solution, I have just tested it and afraid it doesn't work. 

Couple of things, deploy-to-lambda.py needs to include the ISO-639-2_utf-8.txt or it won't be uploaded.

```
if(os.path.isfile("ISO-639-2_utf-8.txt")):
  shutil.copy("ISO-639-2_utf-8.txt", app_name)
```  

Testing the skill (with the txt file uploaded) gives the following error:

```
Syntax error in module 'wsgi': Non-ASCII character '\xc3' in file /var/task/kodi.py on line 987, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details (kodi.py, line 987)
```

I was getting this error on the Flask branch and is why I included `# coding=utf-8`. Sorry if it is not the best practice for defining the encoding type, but I think it is required to handle the non ascii characters. 

(note that I didn't want to open this can of worms, especially for the old skill, so it might be easier (at least till 2.5) that the language response is muted)

jingai:
Instead of punting the problem for the future, let's just fix it here.  The addition to deploy-to-lambda.py is a simple one.

But yes, specifying the character encoding does appear to be required (the pycountry package on my server here does this too).

@m0ngr31, do you know another way around this?  We could always put the offending method in a file by itself if you want, but that doesn't seem to actually help much.

digiltd:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
```

also works

m0ngr31:
> @m0ngr31, do you know another way around this? We could always put the offending method in a file by itself if you want, but that doesn't seem to actually help much.

We'll probably just have to play doing `.encode('utf-8')` on the strings.

jingai:
Yeah.  I'll mess around with it more in a few hours.

m0ngr31:
I pushed my version up to the flask-ask branch based on @digiltd's code. Let me know if you guys run into issues.

jingai:
Why not just rebase or cherry-pick?

Anyway, same problem as on master branch; but I'm looking at it now.

m0ngr31:
Yeah, that's my bad. I was just in the middle of getting rid of the PyBabel dependency and I didn't even think about it.

jingai:
Just worried about a crazy merge back into master if we delay 2.5 much longer lol

m0ngr31:
Yes, I need to get this wrapped up soon. I feel bad for everyone still on the old versions.

jingai:
Ok, so the culprit was in the comment:

```looks up 3 character code in the dictionary e.g. fre|fra|fr|French|français```

Fixing and pushing up in a sec.

jingai:
Should be fixed now.

jingai:
@digiltd lemme know when you've had a chance to test again.  I'll close this then.

jingai:
Believe this is now complete.

PR #131, opened 2017-03-30
Sync streaming music with ListenTo*.

jingai:
This is basically a copy/paste of the existing music methods for streaming.  Ideally, we'll want to factor out the common code, but in the interest of getting 2.5 out the door, we decided to take this route for now.

One 'gotcha' is that I'm having trouble coming up with verbage to allow shuffling playlists by default.  "Alexa, shuffle/play playlist" makes enough sense, but when the main verb here is "stream," I'm not sure how to fire that off.  My suggestion would be to implement AMAZON.Shuffle* (and AMAZON.Loop*) and just make, "Alexa, ask Kodi to stream playlist foo," play it in order by default.  I'll make Issues for these if we decide to go that route.

This is the first(?) PR that affects our translations.  I think the precedent is to insert your own language strings into the other files, which is what I've done.

This does mean that at least as of this PR, a few of the responses and utterances are in English for German deployments as well.  They will remain in English until a kind German-speaking individual can replace them :)

Further, because of the numerous new references to the music slots, I had to reduce the slot generator limit to 400 per slot to get the skill to save.  @m0ngr31 has already taken care of this on the web slot generator.

Finally, we will need to update documentation to warn users that they should re-generate their slot values so they don't encounter the above issue upon upgrading.

### DIFF
--- a/alexa.py
+++ b/alexa.py
@@ -42,7 +42,7 @@ CAN_STREAM = music.has_music_functionality()
 
 # Needs to be instanced after app is configured
 # ask = Ask(app, "/", None, TEMPLATE_FILE) # For when PR gets merged on Flask-Ask project
-ask = Ask(app, '/')
+ask = Ask(app, "/")
 
 
 # Start of intent methods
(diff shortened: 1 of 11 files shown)


jingai:
Could use some additional testers too if anyone's around.

PR #146, opened 2017-04-15
Deep search

jingai:
Instead of simply failing if the skill can't determine the media type the user requested, let's try to search deeper.

If the user specifies the media type, with something like, "Alexa, ask Kodi to play the movie Ghostbusters," then nothing changes.  But if the user says something like, "Alexa, ask Kodi to play/watch Ghostbusters," we should try to search further to find something that matches what they requested.

This PR tries to do that.

It isn't perfect and probably never will be, mostly because Amazon has a foot-hold over the "play" verb.  But, it works well enough for that, and very well if you use the alternate verbs, "watch" and "listen to."

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -35,3 +35,20 @@ SKILL_TZ=
 # Uncomment the following if you'd like "Alexa, tell Kodi to shut down" to
 # quit Kodi instead of shutting down the system.
 # SHUTDOWN_MEANS_QUIT=1
+
+# Uncomment the following to disable deeper searches on generic play commands.
+# By default, if you ask to play media without specifying the type, the skill
+# will search through the entire library to find a match.
+#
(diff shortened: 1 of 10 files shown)


jingai:
For those wanting to test this, bear in mind the prescribed search order.

It's supposed to feel pretty painless for those searching for movies, for instance, as that's the default category; a la Plex.

From there, shows match next (but not episodes), and then musicians, songs, and finally albums.

There is some argument in my own head as to whether I should match songs or artists first.. could use some input there.

Without a curated library database though, this can't be perfect, but it isn't trying to be perfect.  I'm just trying to close the gap.

jingai:
The benefit of matching Artists first is that it's quicker; i.e., there's likely to be fewer artists than songs.

The benefit of matching Songs first is that appears to be Alexa's default behavior for 1st party music.  On a conflict, I've noted that it plays the song.

PR #155, opened 2017-05-05
Added support for mutiple alexa/kodi devices.

ghlynch:
This needs some more work and docs but I would rather get some comments first.

- Added support for multiple alexa devices talking to one or more kodi devices. The typical use case is multiple Kodi in multiple room, each with their own Alexa. Talk in one room and the Kodi in that room responds.
- The environment based config is replaced by a ConfigParser configuration file. See kodi.config as an example. This new config supports sections for each Alexa device (keyed by deviceId) where kodi configuration variable can be overloaded. Defaults apply to all sections/devices if not overloaded.
- Tested with a local server only, not with lambdas. But the device routing is stateless so should work there as well. Need feedback.
- It could be better to key the sections by Kodi device, giving them and abstract name. This would allow dynamic configuration. All Alexas would start with the default Kodi, and an utterance like "use kodi in the office" would switch a device to that kodi. But that requires state, so most likely would need a DB service for lambdas.


ghlynch:
This pr looks broken, let me close and redo.

PR #156, opened 2017-05-05
Added support for mutiple alexa/kodi devices.

ghlynch:
This needs some more work and docs but I would rather get some comments first.

- Added support for multiple alexa devices talking to one or more kodi devices. The typical use case is multiple Kodi in multiple room, each with their own Alexa. Talk in one room and the Kodi in that room responds.
- The environment based config is replaced by a ConfigParser configuration file. See kodi.config as an example. This new config supports sections for each Alexa device (keyed by deviceId) where kodi configuration variable can be overloaded. Defaults apply to all sections/devices if not overloaded.
- Tested with a local server only, not with lambdas. But the device routing is stateless so should work there as well. Need feedback.
- It would probably be better to key the sections by Kodi device, giving them and abstract name. This would allow dynamic configuration. All Alexas would start with the default Kodi, and an utterance like "use kodi in the office" would switch a device to that kodi. But that requires state, so most likely would need a DB service for lambdas.


alexa.py
```diff
@@ -37,6 +33,8 @@
 # Timestamp based verification shouldn't be used in production. Use at own risk
 # app.config['ASK_VERIFY_TIMESTAMP_DEBUG'] = True
 
+# https://github.com/johnwheeler/flask-ask/issues/43
+app.config['ASK_VERIFY_REQUESTS'] = False
```
jingai: I personally don't want to disable this, so at the very least, it should be configurable.

jingai:
Thanks for this!  I will look it over just as soon as I can find some free time to do so.

I had already intended to make a companion addon for Kodi that would allow the user to configure the Echo<->Kodi relationships via voice.  This way the device id could be stored in the Kodi addon config without the user needing a way to look it up on their own.  The addon would go hand-in-hand with what you've done, for the most part.

To make your PR work on its own though, we'll need to devise a way for users to get at the device IDs for their Echos.  I assume you're already aware of this, but just mentioning it here to document it.

> The environment based config is replaced by a ConfigParser configuration file. See kodi.config as an example. This new config supports sections for each Alexa device (keyed by deviceId) where kodi configuration variable can be overloaded. Defaults apply to all sections/devices if not overloaded.

When I first started helping with this project, I moved it to ConfigParser too, but IIRC there was some reason @m0ngr31 wanted to stick with environment variables.  I'll let him chime in about it, but assuming there's no issues with Heroku/AWS/etc I don't have a problem with the change.

At a glance though, I do notice you've missed a few of the configuration items though.  Namely, SKILL_TZ, LANGUAGE, SHUTDOWN_MEANS_QUIT, and DISABLE_DEEP_SEARCH.  As well, SKILL_APPID.. I see why you didn't include this one, as you've disabled this functionality by default, but I'd personally rather not disable it.

Anyway, thanks again.  Both @m0ngr31 and I have become suddenly busy with other things, but we'll try to get to this as soon as we can.

alexa.py
ghlynch: Yes, that was for testing, removed.

ghlynch:
Hello,

Thanks for the quick comments. I addressed them in a commit today.

For now users can pick up their device IDs in the log of the server, I 
added a few comments in the config to explain.

Thanks!

jingai:
I can't review this fully just yet.. but, some thoughts:

* Could you retain the comments from .env.example?  They should map mostly neatly to kodi.config
* kodi.config should be renamed to kodi.config.example so as to not interfere with the user's existing configuration, and kodi.config should be added to .gitignore
* Remove .env.example and from .gitignore
* disable_deep_search needs to accept other boolean values such as 1, 0, true, false, etc
* Why did you remove the try/except and debugging from the call to [requests.post in kodi.py](https://github.com/ghlynch/kodi-alexa/commit/5129d1c18d9a934863a20691e57eb7ab2b24feaa#diff-7f0f68209ca119972d27222e8399d12aL357)?  Printing the actual request in the event of a failure is useful, but if you had a reason for it we can discuss it

I'll go over kodi.py more as soon as I can.  The diff there is a bit harder to read due to the reformatting.  This isn't a problem, just need to find time to do it.

jingai:
Also needs consistent indents -- we use two spaces for each indent.

jingai:
app.json will obviously need some TLC too.

jingai:
(couldn't resist looking at this more...)

I can make up PRs for what I've already noted if you want.  Otherwise it looks really good so far!

ghlynch:
On 5/9/17 12:45 PM, jingai wrote:
>
> I can't review this fully just yet.. but, some thoughts:
>
>   * Could you retain the comments from .env.example? They should map
>     mostly neatly to kodi.config
>
Yes, done.
>
>   * kodi.config should be renamed to kodi.config.example so as to not
>     interfere with the user's existing configuration
>
Done.
>
>   * disable_deep_search needs to accept other boolean values such as
>     1, 0, true, false, etc
>
Ok, I reverted to 1 or blank. Not keen on adding support for multiple 
values for now as I prefer to avoid logic changes until there is a test 
suite.
>
>   * Why did you remove the try/except and debugging from the call to
>     requests.post in kodi.py
>     <https://github.com/ghlynch/kodi-alexa/commit/5129d1c18d9a934863a20691e57eb7ab2b24feaa#diff-7f0f68209ca119972d27222e8399d12aL357>?
>     Printing the actual request in the event of a failure is useful,
>     but if you had a reason for it we can discuss it
>
I think it was my addition on the previous pass. It's not very helpful 
because it traps the request verification errors blindly (r is not 
evaluated in that case by the time the except triggers).
>
> I'll go over kodi.py more as soon as I can. The diff there is a bit 
> harder to read due to the reformatting. This isn't a problem, just 
> need to find time to do it.
>

Hey, I noticed my previous commit generated some comments with more log 
details then I wanted. Would you mind deleting your initial comment? 
from 5/5 7:43 ? I'll take care of rewriting the logs. Sorry about that 
and thanks.

ghlynch:
Oh right.

Who uses that? Looks like Heroku specific documentation?

Separately someone needs to validate that Heroku and Lambda still work 
before merging.

ghlynch:
Cool. I just addressed your first email.

Happy to look at the spacing, need to learn how to do that with Atom 
anyways.

Would help if you can do the app.json as I don't know the context there 
and we're potentially removing "env".

Thanks.

ghlynch:
Oh one more thing: it works really well! Don't know if you have tried it 
yet with multiple devices in multiple rooms, but it is quite amazing.

Cheers and thanks for the original code.

jingai:
> >   * disable_deep_search needs to accept other boolean values such as
> >     1, 0, true, false, etc
> >
> Ok, I reverted to 1 or blank. Not keen on adding support for multiple values for now as I prefer to avoid logic changes until there is a test suite.

I'll need to look again, but I believe there's a way to test for all reasonable possibilities built into ConfigParser.

It's not that I had an issue with you using "yes/no," but that it should accept all other representations of true/false as well.

> >   * Why did you remove the try/except and debugging from the call to
> >     requests.post in kodi.py
> >     <https://github.com/ghlynch/kodi-alexa/commit/5129d1c18d9a934863a20691e57eb7ab2b24feaa#diff-7f0f68209ca119972d27222e8399d12aL357>?
> >     Printing the actual request in the event of a failure is useful,
> >     but if you had a reason for it we can discuss it
>
> I think it was my addition on the previous pass. It's not very helpful because it traps the request verification errors blindly (r is not evaluated in that case by the time the except triggers).

Ah that is correct.  Would be useful to print the request itself though if the json.loads() call fails.  Printing the request on failure, however it happens, is useful in generating a similar request by hand to debug.

> Separately someone needs to validate that Heroku and Lambda still work before merging.
>
> Would help if you can do the app.json as I don't know the context there and we're potentially removing "env".

I host the skill locally too, but @m0ngr31 or @digiltd can test it with those.

> Oh one more thing: it works really well! Don't know if you have tried it yet with multiple devices in multiple rooms, but it is quite amazing.

I haven't actually tried it yet -- I've only just now fetched your changes here to look over.  I like to know what I'm trying before I do so :)

ghlynch:
Great, let me know if you have any issues.

jingai:
I'll check after dinner.  Go ahead and push up your latest changes if you could.

jingai:
I've set aside a few hours now to get this ready to go in, but I'm not sure what you've already done and are currently working on.  If you could let me know we can probably bang out the remaining loose ends pretty quickly.

ghlynch:
It's actually working when you serve locally. So you just need to set up 
your kodi.config an run. I use:

gunicorn -w3 --certfile=server-crt.pem --keyfile=server-key.pem -b :8000 
alexa:app

The device ID is printed on stdout. The first request from a device with 
an ID will fail unless you setup DEFAULT to point to a valid Kodi.

LMK if you have any trouble.

jingai:
I understand that it works, but I'm looking to polish it up to include this in mainline.  You said you've already made some changes that I commented on, but I don't yet see them in your repo.

ghlynch:
I see. My bad probably. Let me check now.

ghlynch:
Pushed.

jingai:
Thanks.  How long will you be around today?  I'm going to get some PRs going soon.

ghlynch:
I can take a look later tonight and also tomorrow morning.

jingai:
I think the biggest hurdle you're going to face with this is the switch from environment variables to ConfigParser for the configuration.  I don't personally care one way or the other, but doing this does break @m0ngr31 's "one-button-deploy-to-heroku" functionality you might have seen in the README.

If he can devise a way to resurrect that with the new configuration file, it'll probably be OK, but I don't know enough about Heroku personally to know what can and can't be done here.

That said, from what I can tell, here's what's left to make this worthy of a merge:

* Needs full testing since this basically touches everything, albeit lightly
* Rebase onto current master (I can handle this at the end)

jingai:
@m0ngr31, once this is complete, I vote that it qualifies as a point release (2.6), primarily due to the configuration changes.

I'm personally undecided as to whether we should actually document the ability to configure the Kodi<->Alexa relationships at this point, as I think it will cause a lot of support issues.  But, doing the migration now allows us to get people moved over to the new configuration file without adding more complexity for them; that is, they do the grunt work for that hurdle and everything stays the same from their perspective.  For the more savvy users, they'll probably notice the example device configurations in kodi.config.example and this allows them to utilize it immediately.

In the meantime, I can work on the Kodi addon script that can handle storing the device IDs.

The only real issue -- besides testing -- is what to do with app.json and the one-button Heroku deployment.

ghlynch:
On 5/9/17 6:38 PM, jingai wrote:
>
> I think the biggest hurdle you're going to face with this is the 
> switch from environment variables to ConfigParser for the 
> configuration. I don't personally care one way or the other, but doing 
> this does break @m0ngr31 <https://github.com/m0ngr31> 's 
> "one-button-deploy-to-heroku" functionality you might have seen in the 
> README.
>
> If he can devise a way to resurrect that with the new configuration 
> file, it'll probably be OK, but I don't know enough about Heroku 
> personally to know what can and can't be done here.
>
Makes sense. I don't know Heroku enough either, but if someone can point 
in the right direction happy to help.
>
> That said, from what I can tell, here's what's left to make this 
> worthy of a merge:
>
>   * Documentation needs updating to point the users at kodi.config
>     instead of .env
>   * Needs full testing since this basically touches everything, albeit
>     lightly
>
Local testing works. I can probably test lambda, and try to add a basis 
test suite as well within the next week or so.

jingai:
From talking to @m0ngr31 , Heroku only really supports environment variables for configuration.  If we're to continue supporting Heroku deployments, we'll have to think of something.

m0ngr31:
I made a PR for this branch to allow Heroku users to continue to use the skill the same way they are currently.

m0ngr31:
I made another PR, this time replacing the kodi.py file with the new `kodi-voice` library and switching back an option I commented on a little earlier.

jingai:
Moved this to upstream branch version-2.6 to ready it for release.

PR #160, opened 2017-05-12
Version 2.6

jingai:
Readying for release.


jingai:
I believe the only thing left is to describe how to set up the multi-device support.  The README should direct users of AWS Lambda and Heroku on how to look at their logs to identify the device IDs.

PR #181, opened 2017-06-01
Reduce the max json results when requesting latest movies to 10.

Josh5:
This more than halves the Alexa skill response time and makes for a better result for the user on what NEW movies he/she has. Opposed to sending back 90 results including movies that may have been added a year ago but never bothered watching.

### DIFF
--- a/alexa.py
+++ b/alexa.py
@@ -2088,9 +2088,9 @@ def alexa_what_new_movies(Genre):
 
   # Select from specified genre if one was matched
   if genre_located:
-    new_movies = kodi.GetUnwatchedMoviesByGenre(genre_located['label'])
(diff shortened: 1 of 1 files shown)


jingai:
The number of results should just be configurable. I personally want all of them.

Josh5:
So you would want a config option for setting a number of results inside of kodi.config?
Could add another section to it or just chuck it under "[DEFAULT]". What do you think?

jingai:
Probably under global. I'll add it tomorrow and just handle it all in Kodi-Voice.

Josh5:
Please see this pull request for a configurable option:
https://github.com/m0ngr31/kodi-voice/pull/6

PR #182, opened 2017-06-02
Remove OpenRemote Intent.

jingai:
Instead, all navigation commands now by default leave the session open.

Nearly every time I've used one of the nav commands, I've followed it with more.  To me, it makes sense to just leave the session open whenever a nav command is issued.

Further, the user can also begin a session by simply opening the skill with no command ("open kodi"), and with the above, the session will stay open if it receives a navigation command first, so there doesn't seem to be much need for an explicit "open remote" intent.

### DIFF
--- a/alexa.py
+++ b/alexa.py
@@ -1138,9 +1138,8 @@ def alexa_player_move_up():
 
   kodi = Kodi(config, context)
   kodi.PlayerMoveUp()
-  response_text = ""
-
-  return statement(response_text).simple_card(card_title, response_text)
+  response_text = render_template('confirmed').encode("utf-8")
+  return question(response_text)
 
 
 # Handle the PlayerMoveDown intent.
@@ -1151,9 +1150,8 @@ def alexa_player_move_down():
 
   kodi = Kodi(config, context)
   kodi.PlayerMoveDown()
-  response_text = ""
-
-  return statement(response_text).simple_card(card_title, response_text)
+  response_text = render_template('confirmed').encode("utf-8")
+  return question(response_text)
 
 
 # Handle the PlayerMoveLeft intent.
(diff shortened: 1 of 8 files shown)


m0ngr31:
Looks fine to me. Just trying to whittle down some intents?

jingai:
No, I just found I never used navigation commands because it took two steps to open a session.  Then, I thought about it some.. and never really use just one nav command, so... :)

jingai:
Also, "open kodi and open remote" seemed kind of silly, since "open kodi" already leaves it open.

jingai:
Hmm, actually.. tell me what you think of this:  it occurs to me that the What* Intents maybe should also leave the session open, _if_ they were initiated from an already open session.  So, the dialog could go like:

```
Alexa, open Kodi
What new movies do I have?
Play movie_she_just_told_me_about
```

Thoughts?

m0ngr31:
Yeah, I like that a lot.

jingai:
Ok, lemme implement that and I'll update here.

jingai:
How about that?

jingai:
btw, changing the translation for `confirmed` to `<speak><say-as interpret-as="interjection">bam</say-as></speak>` is kinda.. fun lol

jingai:
Yeah, I have to say, it feels a lot more natural being able to follow up queries with playback commands.

jingai:
For reference, I limited the definition of _query_ in this context to requests that:

1. Ask Kodi for information from its library, and
2. Don't perform any visible action in Kodi.*

\* with the sole exception of AddonGlobalSearch, because it seemed logical to follow that up with a navigation command or a playback request.

Navigation-type requests just unconditionally keep the session open.

PR #183, opened 2017-06-03
Add basic PVR support: finding channels and programmes

ta264:
Based on #47 by @freemans13.  Also depends on the PR in kodi-voice
https://github.com/m0ngr31/kodi-voice/pull/7

### DIFF
--- a/generate_custom_slots.py
+++ b/generate_custom_slots.py
@@ -215,3 +215,42 @@ gfile = open('ADDONS', 'w')
 for a in cleaned:
   gfile.write("%s\n" % a.encode("utf-8"))
 gfile.close()
+
+# Generetate CHANNELS Slot
+retrieved = kodi.GetPVRChannels()
+
+all = []
+
+if 'result' in retrieved and 'channels' in retrieved['result']:
+  for v in retrieved['result']['channels']:
+    ascii_name = v['sanitized_label'].encode('ascii', 'replace')
+    all.append(ascii_name.encode('utf-8').strip())
+
+deduped = list(set(all))
+
+gfile = open('CHANNELS', 'w')
+for a in deduped:
+  gfile.write("%s\n" % a)
+gfile.close()
+
+# Generetate BROADCASTS Slot
+retrieved = kodi.GetPVRChannels()
+      
+all = []
+      
+if 'result' in retrieved and 'channels' in retrieved['result']:
+  for v in retrieved['result']['channels']:
+    channelid = v['channelid']
+    broadcasts = kodi.GetPVRBroadcasts(channelid)
+    if 'result' in broadcasts and 'broadcasts' in broadcasts['result']:
+      for broadcast in broadcasts['result']['broadcasts']:
+        ascii_name = broadcast['label'].encode('ascii', 'replace')
+        removed_paren = re.sub(r'\([^)]*\)', '', ascii_name).rstrip().lower().translate(None, string.punctuation)
+        all.append(removed_paren.encode('utf-8').strip())
+
+deduped = list(set(all))
+
+gfile = open('BROADCASTS', 'w')
+for a in deduped:
+  gfile.write("%s\n" % a)
+gfile.close()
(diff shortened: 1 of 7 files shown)


jingai:
Thanks @ta264.

My first thought is that `find` and `look for` don't fit with the rest of the design.

mcl22:
Just a thought, perhaps we should add an intend to open the epg, too. So one can Alexa have open the epg and then tell her to play channel whatever.

ta264:
@jingai I agree.  I didn't really pay much attention to the utterances.  Do you think this would be better:

WatchPVRBroadcast (play/watch) (/the) (/t.v./live t.v.) (program/broadcast) {Broadcast}
WatchPVRBroadcast change channel to {Broadcast}
WatchPVRChannel (play/watch) (/the) (p.v.r./t.v./live t.v.) (/channel) {Channel}
WatchPVRChannel (play/watch) (/the) channel {Channel}

jingai:
Yes, I think that fits better.  Just FYI though, in the sample utterances, Amazon's docs say:

`Acronyms and other phrases involving spoken letters should be separated by periods and spaces (“n. b. a.”, not “nba”).`

So, it should be, for example, `p. v. r.` and not `p.v.r.`.

generate_custom_slots.py
```diff
@@ -215,3 +215,42 @@
...
 gfile.close()
+
+# Generetate CHANNELS Slot
+retrieved = kodi.GetPVRChannels()
+
+all = []
+
+if 'result' in retrieved and 'channels' in retrieved['result']:
+  for v in retrieved['result']['channels']:
+    ascii_name = v['sanitized_label'].encode('ascii', 'replace')
```
jingai, line 226: Should this be ASCII-only?  We support UTF-8 in the other slots.

You also still need to call kodi.sanitize_name() (or a subset of it, if it doesn't all apply to channels) because there are characters that Amazon restricts in slot items.

jingai:
I know this suggestion might not be taken well, but in my opinion the PVR stuff should be a separate skill.

My reasoning is that there's a lot of overlap, and some of it I think people would want.  For instance, if it were a separate skill, something like:

`Alexa, ask PVR to watch The Office`

could be made to look through current and recorded programs and just switch to the right channel/recording.

Since the core communication code has been broken out to kodi-voice, I think this isn't an unreasonable thing to do.

Further, the skill has grown to a point that we're having trouble keeping it within Amazon's own size requirements.  I can envision that as we add more and more PVR functions, we're going to end up with issues.

For reference, we already had to give up on integrating streaming music support in the skill due to this.  PVR support will end up with very similar problems, IMHO.

Honestly, it would end up being much easier to make it a separate skill anyway. Having most of the core code in a library makes it much easier to make a new skill, plus you won't constantly be fighting collisions with all of the other library intents we already have :)

jingai:
If you want to have a go at making a new skill for PVR, I will certainly help you along.  I don't actually have a PVR, but I have a good understanding of how the skill works and how things should work for PVR.

I would be able to do some rudimentary testing with the dummy PVR addon too, I suppose.

danielmasur:
I can test it and also do the German translations. I have my VU+ Box hooked
up via the VU+ Client.

Am 04.06.2017 20:52 schrieb "jingai" <notifications@github.com>:

mateuszdrab:
Hey guys

Do we have an update on this? The major reason I use Kodi is for PVR access and this pull request seems to have what I need to make it work.

I think I'll download the main branch and merge this pull request manually and give it a go.

Anyone using this code in their environment?

Regards

jingai:
@mateuszdrab, I still believe this should be its own skill, which means some one just needs to take the time to write it.

im85288:
I tried to merge this into my fork but had no success getting it to actually start playback of a channel. It’s a shame this has not been merged as for people with a pvr it is more used than other parts such as movies/tvshows etc.

When I get more time I may try and get this into its own skill but the fact that the above did not work makes me feel it may not be ready at this moment in time

jingai:
@im85288 it could definitely work, but it needs to be its own skill.  The PR is left open just to provide a jumping-off point for anyone that wants to write the skill.

m0ngr31:
Closing this. Been open too long.

PR #189, opened 2017-06-10
Replace all calls to Kodi.matchHeard().

jingai:
Stop directly calling Kodi.matchHeard() to better abstract things.  Instead, utilize the Find*() helper methods in Kodi-Voice.

This supports the trial of two methods of handling matching:

1. Kodi.matchHeard() being been moved to [script.alexa.helper](https://github.com/jingai/script.alexa.helper).  This way removes the need to transmit the library items to the skill, at the cost of off-loading the fuzzy matching to the Kodi client machine.  The addon must be installed on your Kodi clients for this to work.  This method utilizes the changes in m0ngr31/kodi-voice/pull/11

2. Optimizing matching and library item transmission in Kodi-Voice.  This is the old method, and has the benefit of potentially placing the fuzzy matching job on a more powerful machine, at the expense of having to transmit the list items.  This method utilizes the changes in _FORTHCOMING_PR_.

### DIFF
--- a/alexa.py
+++ b/alexa.py
@@ -630,44 +630,36 @@ def alexa_listen_album(Album, Artist, shuffle=False):
   print card_title
 
   kodi = Kodi(config, context)
+
+  album_id = None
   if Artist:
     artist_id, artist_label = kodi.FindArtist(Artist)
     if artist_id:
-      albums = kodi.GetArtistAlbums(artist_id)
-      if 'result' in albums and 'albums' in albums['result']:
-        albums_list = albums['result']['albums']
-        album_located = kodi.matchHeard(Album, albums_list, 'label')
-
-        if album_located:
-          album_id = album_located['albumid']
-          album_label = album_located['label']
-          kodi.PlayerStop()
-          kodi.ClearAudioPlaylist()
-          kodi.AddAlbumToPlaylist(album_id, shuffle)
-          kodi.StartAudioPlaylist()
-          if shuffle:
-            response_text = render_template('shuffling_album_artist', album_name=album_label, artist=artist_label).encode("utf-8")
-          else:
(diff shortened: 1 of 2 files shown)


jingai:
Tested so far on x86_64 linux, android arm7, and x86_64 darwin.  No issues, besides being a bit slow if python-Levenshtein isn't available.

On Linux, Kodi apparently will happily search system path for libraries, so I get it for 'free' there.  On macOS, it doesn't do this.. so I manually copied it into resource/lib/ just to see, and it definitely makes a noticeable difference on more underpowered/older hardware.

On my Fire TV Stick.. it's slow.  Probably not slower than it was before if you host the skill on AWS/Heroku, but it's most definitely slower for me than doing the matching server-side since my server is a much beefier machine and everything is on the LAN here.

Shipping the addon with support for pre-compiled Levenshtein libraries for all of the Kodi-supported platforms is doable.  I just don't have an Android development environment atm.

I'm going to do some profiling on my oldest Mac mini and see if I can identify any other areas of improvement.

digiltd:
Seems to work nicely when I tested it on my Mac mini (late 2012, 16gb)

Well done, it will make a big difference not having to bounce the data back and forth all the time

🏆

jingai:
@digiltd, I'm going to need some hard numbers from you guys to really figure out exactly where the bottle-neck is.  The addon approach works fine if the Kodi client is on a sufficiently-powerful computer, but on embedded devices (like the Fire TV [stick]), it's very, very slow.  So I have some things I want to try.

If you could.. I'd like to know how long it takes (down to the millisecond please) with the addon on your setup to execute `play album`, where _album_ is some album you have, preferably one that would be difficult to match directly.  Execute this a few times please.

Logs from Kodi would be helpful too -- I only need the output from the Alexa Helper addon.

jingai:
Moved to #190 as this will support two different trials.

PR #200, opened 2017-07-10
send text to kodi, inc DE translation

digiltd:
needs the modified kodi-voice (https://github.com/digiltd/kodi-voice/tree/send-text)

you say "enter the text whatever you fancy"

she replies "I heard the text: whatever you fancy, is this correct?"

"yes" sends it

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,8 +156,11 @@ You need to create 9 different slots:
 - MUSICPLAYLISTS
 - VIDEOPLAYLISTS
 - ADDONS
+- GENERICPHRASES
 
 To make it as easy as possible, I wrote a little webapp that will give you the information you need: [here](https://slot-generator.herokuapp.com/).
+
+For "GENERICPHRASES" (used to send text to Kodi) use the sample data included the speech_assets folder.
 
 You can also get the information from running `python generate_custom_slots.py` in the project directory. This will create txt files with the relevant information. This will communicate with the Kodi instance defined in the `[DEFAULT]` section in `kodi.config`. *NOTE: If you're deploying to Heroku, you cannot use the python slot generator unless you create the kodi.config file.  It's best to use the webapp for generating slots in this case.*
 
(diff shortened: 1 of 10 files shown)


jingai:
Have had some discussion on Slack that probably should go here.

Basically, the main concerns are:

* Things that aren't natural phrases, such as URLs, IP addresses, etc.  These come through Alexa in ways that we'd need to either accept will likely never be what the user intended, or would need translation in the skill.  This is a complex problem since we'd need to identify the user's intention.  Further, the problem is compounded by the fact that we also support German (and I'm assuming other languages in the future)

* Collisions with other utterances, since it's more likely that the slot value here might sound similar to one of our other utterances.  We were able to get her to fire the wrong Intent by asking for a URL and oddly, the phrase, `i have got a lovely bunch of coconuts`.  I'm inclined to think we probably will find more, and testing for that is difficult.

jingai:
I've put up PR https://github.com/digiltd/kodi-alexa/pull/1 for changes that I think should go into this as well.

mcl22:
Here are the translations (I found in the code):

EnterText schreibe (/den) Text {Phrase}
EnterText gib (den/folgenden) Text ein {Phrase}  

Ich habe folgenden Text gehört:
{{ heard_text }}
Ist das korrekt?
 
Du möchtest dass ich folgenden Text eingebe:
{{ heard_text }}
Richtig?

digiltd:
@mcl22 not sure it is going in or not, but can you also do your magic for:

try_enter_text_again: "Try again, say: Enter Text. followed by the text you want to enter."

mcl22:
Abracadabra ... 
try_enter_text_again: "Versuche es erneut, sage: gib Text ein, gefolgt von dem Text, den du eingeben möchtest."

digiltd:
Thanks, I have updated it at my end, the try_enter_text_again question is a bit of a mouthful in both languages. Ideally she would ask "you said blah blah, if this correct?" and if you said "no" she would just say "say it again". But when she starts listening again, you can trigger any intent, so it is important to make sure the users says the whole phrase.

Issue #49, opened 2016-10-26
Subtitles can be enabled/disabled but not downloaded


x5nder:
Don't know where to suggest this, so I hope it's okay I'm posting it here:

I am missing the option to download a subtitle. If I'm not mistaken, the code would be:

def SubtitlesGet():
  playerid = GetVideoPlayerID()
  if playerid:
    return SendCommand(RPCString("GUI.ActivateWindow", {"window":"subtitlesearch"}))

with alexa.intents:
    {
      "intent": "SubtitlesGet",
      "slots": []
    },

and alexa.utterances:
SubtitlesGet download subtitles
SubtitlesGet get subtitles
SubtitlesGet find subtitles

(Since I have no GitHub experience, I am definitely not trying to implement this myself, haha)

jingai:
Thanks for the examples.  I'll have a look at this probably this weekend.

m0ngr31:
I have this added, but it doesn't show anything for me. I don't use subtitles, so I'm guessing I need to do some setup to enable the search to show up?

jingai:
Can you activate the subtitle search in Kodi using remote/keyboard?

fonzie24:
There is no specific shortcut key to open the subtitle search that I am aware of.  However, you can customize the keymap to create the shortcut.

"subtitlesearch" Window ID is 10153 and it points to the dialoguesubtitles.xml

http://kodi.wiki/view/keymap

jingai:
I'm kind of inclined to say this is out of scope for the Alexa skill.  It requires a lot of interaction after invoking that honestly is best handled with a remote/keyboard/mouse.

It would be possible to just download the first available via voice, but it's probably not likely to be what the user wanted.  Unsure, honestly..

jingai:
I'm inclined to think that for a VUI, it should just download the first hit for the user's locale rather than popping a dialog for choosing what to download.

But I'd like some input from others on this before I tackle the problem.

fonzie24:
Yes, I'd be fine with it downloading the first subtitle file from the list of the subtitle provider that you have set as default.

jingai:
I'll probably push forward with that then and see if there are any complaints.

fonzie24:
sounds good. thanks for keeping us updated.

jingai:
Just curious, but what's wrong with letting Kodi just auto-download the first available subtitle for your language for this?

[...]

Issue #16, opened 2016-10-05
Application.Quit


jingai:
Unable to add a Quit intent to quit Kodi, because Amazon intercepts quit, exit, and close as requests to end the session.

Unless someone can think of an alternate way to invoke this, we'll probably have to settle for just Shutdown (which can be mapped in Kodi to Quit).

jingai:
Hmm, apparently Kodi doesn't respect the user's configured action for Shutdown when called via JSON RPC.  Annoying.

jingai:
I think the problem is that the "quit" utterance is reserved to tell the skill itself to quit (i.e., SessionEndedRequest).  We could intercept this under certain conditions, like maybe if 'reason' is 'USER_INITIATED' and quit Kodi, but once we start supporting keeping sessions open, this might lead to accidentally quitting Kodi if the user is just trying to tell the skill to stop talking to them.

Of course, we could just check to see if the session is currently open and map it accordingly.  Would have to think it through is all.

Another approach would be to make the action of the shutdown intent configurable via an environment variable.  On my local branch I currently map shut down to quit for this reason, but it'd be easy enough to make it configurable so all could use it if they choose to.

Anybody have any particular preference?

fonzie24:
I notice you have many power commands, such as: "Hibernate, Reboot, Shutdown, Suspend"

However, I was wondering if it would be possible to add an intent to log off a user profile?

I tried to do it myself, but I couldn't get it to work.

I have multiple user accounts and it would be helpful to have a voice command to quickly log off the current user to make sure the watched statuses get synched correctly (among other things).

jingai:
@fonzie24 , open a separate Issue please and we'll see what we can do.

jingai:
I'm probably going to make this configurable via an environment variable unless anyone objects.

jingai:
Just reminding myself to take care of #17 at the same time.

PR #205, opened 2017-07-23
Add DownloadSubtitles intent

rinzeb:
Added an intent for downloading subtitles. When playing a show or movie, this will open the "Download subtitles" window, where you can download a subtitle from one of your enabled subtitle addons. I prefer to enable the Kodi setting "Automatically download the first subtitle" in order to download subtitles with just 1 voice command. Alternatively, this requires a "Select" command to download a subtitle.
(Related to issue #49 )

### DIFF
--- a/speech_assets/SampleUtterances.german.txt
+++ b/speech_assets/SampleUtterances.german.txt
@@ -1,3 +1,8 @@
+# A list of utterance variations written in a shorthand syntax
+# Syntax and information
+# Tool to expand the single line into the usable list of utterances
+# http://www.makermusings.com/amazon-echo-utterance-expander/
+# http://www.makermusings.com/two thousand fifteen/seven/twenty one/defining-utterances-for-the-alexa-skills-kit/
 AddonExecute führe add on {Addon} aus
 AddonExecute führe addon {Addon} aus
 AddonExecute führe erweiterung {Addon} aus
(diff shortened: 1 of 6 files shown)


rinzeb:
Requires a PR on kodi-voice too: https://github.com/m0ngr31/kodi-voice/pull/16

speech_assets/SampleUtterances.german.txt
```diff
@@ -1,3 +1,8 @@
+# A list of utterance variations written in a shorthand syntax
+# Syntax and information
+# Tool to expand the single line into the usable list of utterances
+# http://www.makermusings.com/amazon-echo-utterance-expander/
+# http://www.makermusings.com/two thousand fifteen/seven/twenty one/defining-utterances-for-the-alexa-skills-kit/
```
jingai, line 5: Why did you add this comment here?  This isn't the file containing the shorthand versions.

jingai:
Thanks for the PR.

I don't think we should use "search for" here though, since that's how we invoke the Global Search addon.

jingai:
Should this return `question()`?  If the auto-download-first-available option is enabled, it doesn't pop a dialog, but there's not really any way for the skill to know that.

jingai:
Merged with the above changes.  Thanks again!

Issue #17, opened 2016-10-05
Destructive commands should require confirmation


jingai:
Hibernate, Reboot, Shutdown, Suspend.

Since they can be destructive (loss of user's work in other applications), Alexa should probably prompt with "Are you sure you want to do this?".  But this requires keeping track of the session, which we don't currently do.

m0ngr31:
I just commited some code to play with saving the session data, so this is something we can start looking at.

jingai:
I'll fool around with it more after we've merged the flask-ask branch.  I want [Quit](https://github.com/m0ngr31/kodi-alexa/issues/16) to work too, which is somewhat related.  Currently I just maintain local changes to make Shutdown function as Quit.

PR #206, opened 2017-07-29
added ViewVideoPlaylist

digiltd:
opens the found video playlist in the GUI

requires the updated kodi-voice: https://github.com/m0ngr31/kodi-voice/pull/17 

I cobbled together a DE translation based on the exisiting ViewMoviews, ViewShows

ViewVideoPlaylist (öffne/gehe zu) video playlist {VideoPlaylist}
ViewVideoPlaylist (öffne/gehe zu) {VideoPlaylist} video playlist

Though need a translation for the template

showing_empty: "Showing"

### DIFF
--- a/alexa.py
+++ b/alexa.py
@@ -2024,9 +2024,11 @@ def alexa_watch_last_show():
 
 # Handle the WatchVideoPlaylist intent.
 @ask.intent('WatchVideoPlaylist')
-def alexa_watch_video_playlist(VideoPlaylist, shuffle=False):
+def alexa_watch_video_playlist(VideoPlaylist, shuffle=False, show=False):
   if shuffle:
     op = render_template('shuffling_empty').encode("utf-8")
+  elif show:
+    op = render_template('showing_empty').encode("utf-8")
   else:
     op = render_template('playing_empty').encode("utf-8")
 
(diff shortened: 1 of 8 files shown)


jingai:
I think this should be self-contained rather than recycling alexa_watch_video_playlist(), like the others are.  Since it's a nav command, there shouldn't be a response either (it should return `question(' ')`).

Why did the German utterance file change so much?  I ran the them through the utterance generator to check, and nothing else changed like it did in this PR.

Lastly, we'll need to add at the same time support for audio playlists and a 'generic' phrasing that excludes the playlist type to line up with the other commands (i.e., `open playlist foo`).  This needs to happen simultaneously to ensure there aren't collisions with each other or the other Intents that are similar.

We'll need to test for weighting issues as well with shorter open commands (`open movies` vs `open movie playlist`, etc), since it's possible Alexa might favor these new commands much more due to the slots involved, which equates to more sample phrases.

jingai:
If we add much more for `open` here, we might also want to consider removing `open` as an invocation word for `AddonExecute`.

jingai:
Superseded by https://github.com/m0ngr31/kodi-alexa/pull/207 which includes what I talked about above.

PR #207, opened 2017-07-31
Add View(Audio|Video)Playlist Intents.

jingai:
Opens (instead of playing) the playlist specified by the user, including the various built-in Recently Added playlists.

German translations for the new utterances need verification.  Ping @mcl22 and @ausweider.


jingai:
German utterances added with 5586681, but I just guessed.  @mcl22 and/or @ausweider should verify before merge.

jingai:
Tested in English and working OK.  Going to merge -- German users should look over the translations and make a PR or comment here if something is wrong.

PR #224, opened 2017-09-17
Dockerfile for building kodi-alexa docker image

mboeru:


jingai:
Thanks!

Two things:

1) It needs to instruct the user how to clone the repo (as it does in the AWS Lambda section),
2) Document how to use kodi.config instead of env vars.

For 2, I realize your preference is env vars, but I'd rather limit that to Heroku since a Docker deployment can support mapping devices.

jingai:
Oh, we also need some text in UPGRADING to describe what they need to do to upgrade to later versions.

m0ngr31:
https://github.com/linuxserver/docker-kanzi

PR #235, opened 2017-11-28
Ready for media-center.io skill

m0ngr31:


jingai:
Looks fine -- I'll test it this weekend to make sure it doesn't break anything for us self-hosters :)

PR #90, opened 2017-01-04
small tweaks to the utterances

digiltd:
open for review of course

I changed the "open" remote to:

`OpenRemote (/start/activate/enable) (navigation mode/navigation/button mode/buttons/remote/remote control(/s)/controls/controller)`

I think we can trim down some of the "new shows/movies/tracks" and "time remaining" utterances as in the case of WhatNewMovies there are 816 utterances :)

PR #241, opened 2017-12-11
a few more utterances for WhatNewEpisodes and WatchEpisode

digiltd:
after a test of "any new episodes of the show" started playing random episodes of the show

### DIFF
--- a/speech_assets/SampleUtterances.en.txt
+++ b/speech_assets/SampleUtterances.en.txt
@@ -1479,6 +1479,9 @@ WhatNewEpisodes if we have a new episode of {Show}
 WhatNewEpisodes if we have an episode of {Show}
 WhatNewEpisodes if we have any episodes of {Show}
 WhatNewEpisodes if we have any new episodes of {Show}
+WhatNewEpisodes is there a new episode of {Show}
+WhatNewEpisodes is there any new episodes of {Show}
+WhatNewEpisodes any new episodes of {Show}
 WhatNewMovies are there any new {MovieGenre} films
 WhatNewMovies are there any new {MovieGenre} films to watch
 WhatNewMovies are there any new {MovieGenre} movies
--- a/speech_assets/SampleUtterances.en.txt
+++ b/speech_assets/SampleUtterances.en.txt
@@ -1318,10 +1318,22 @@ VolumeUp turn up volume
 VolumeUp turn volume up
 VolumeUp up volume
 VolumeUp volume up
+WatchEpisode play episode {Episode} season {Season} of {Show}
+WatchEpisode play episode {Episode} series {Season} of {Show}
 WatchEpisode play season {Season} episode {Episode} of {Show}
 WatchEpisode play series {Season} episode {Episode} of {Show}
+WatchEpisode play {Show} episode {Episode} season {Season}
+WatchEpisode play {Show} episode {Episode} series {Season}
+WatchEpisode play {Show} season {Season} episode {Episode}
+WatchEpisode play {Show} series {Season} episode {Episode}
+WatchEpisode watch episode {Episode} season {Season} of {Show}
+WatchEpisode watch episode {Episode} series {Season} of {Show}
 WatchEpisode watch season {Season} episode {Episode} of {Show}
 WatchEpisode watch series {Season} episode {Episode} of {Show}
+WatchEpisode watch {Show} episode {Episode} season {Season}
+WatchEpisode watch {Show} episode {Episode} series {Season}
+WatchEpisode watch {Show} season {Season} episode {Episode}
+WatchEpisode watch {Show} series {Season} episode {Episode}
 WatchLastShow continue last show
 WatchLastShow continue playing last show
 WatchLastShow continue playing show
--- a/speech_assets/SampleUtterances.en.txt
+++ b/speech_assets/SampleUtterances.en.txt
@@ -1391,6 +1391,10 @@ WatchRandomEpisode play a random episode of {Show}
 WatchRandomEpisode play random episode of {Show}
 WatchRandomEpisode watch a random episode of {Show}
 WatchRandomEpisode watch random episode of {Show}
+WatchRandomEpisode play a random {Show} episode
+WatchRandomEpisode play random {Show} episode
+WatchRandomEpisode watch a random {Show} episode
+WatchRandomEpisode watch random {Show} episode
 WatchRandomMovie play a random film
 WatchRandomMovie play a random movie
 WatchRandomMovie play a random {MovieGenre} film
--- a/speech_assets/SampleUtterances.en.txt
+++ b/speech_assets/SampleUtterances.en.txt
@@ -1480,6 +1480,9 @@ WhatNewEpisodes if we have an episode of {Show}
 WhatNewEpisodes if we have any episodes of {Show}
 WhatNewEpisodes if we have any new episodes of {Show}
 WhatNewEpisodes is there a new episode of {Show}
+WhatNewEpisodes are there any new episodes of {Show}
+WhatNewEpisodes are there new episodes of {Show}
+WhatNewEpisodes is there new episodes of {Show}
 WhatNewEpisodes is there any new episodes of {Show}
 WhatNewEpisodes any new episodes of {Show}
 WhatNewMovies are there any new {MovieGenre} films
--- a/utterances.en.txt
+++ b/utterances.en.txt
@@ -152,6 +152,11 @@ WatchEpisode (play/watch) {Show} episode {Episode} (season/series) {Season}
 
 WatchRandomEpisode (play/watch) (/a) random episode of {Show}
 
+WatchRandomEpisode play a random {Show} episode
+WatchRandomEpisode play random {Show} episode
+WatchRandomEpisode watch a random {Show} episode
+WatchRandomEpisode watch random {Show} episode
+
 WatchRandomShow (play/watch) (/a) random (/{ShowGenre}) (/t. v.) show
 
 WatchNextEpisode (play/watch) (/the) next episode of {Show}
--- a/utterances.en.txt
+++ b/utterances.en.txt
@@ -146,6 +146,9 @@ WatchRandomMovie (play/watch) (/a) random (/{MovieGenre}) (movie/film)
 WatchMovieTrailer (play/watch) (/the) (trailer/preview) (/for {Movie})
 
 WatchEpisode (play/watch) (season/series) {Season} episode {Episode} of {Show}
+WatchEpisode (play/watch) episode {Episode} (season/series) {Season} of {Show}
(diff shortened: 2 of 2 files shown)


jingai:
`is there any new episodes of {Show}` isn't exactly proper grammar, but I suppose someone could say it lol.

Looks fine to me.

digiltd:
I expect many do :)

"is there any" and "are there any" are both valid .... in fact, that got me thinking and I have just added a couple more

m0ngr31:
> after a test of "any new episodes of the show" started playing random episodes of the show

I actually have the exact opposite problem. It's always telling me how many new episodes I have instead of playing a random one.

digiltd:
added some for WatchEpisode to fix #242 (https://forum.kodi.tv/showthread.php?tid=254502&pid=2678453#pid2678453)

A bit deja vu as I was sure I had added those in before. I did, back in Jan #90

jingai:
From the forum post, I believe in #242 the user was trying to do something like, `Alexa, ask Kodi to play The Simpsons`.  This is supposed to work (it's handled in the code), but Alexa often doesn't match such a short invocation well enough, so it falls through -- in this case, it falls through to the handler for recently added songs.  In the case of #242, he/she doesn't have any songs apparently, thus the error.

Your additions for `WatchEpisode` are fine though, provided they don't screw with anything else.  I do remember playing with variations on that phrasing and I can't remember why I decided to omit the alternates.  It's worth some testing, IMHO.

FWIW, these all work for me without adding them explicitly:
```
WhatNewEpisodes is there a new episode of {Show}
WhatNewEpisodes are there any new episodes of {Show}
WhatNewEpisodes are there new episodes of {Show}
```

This works as well without adding it, though again, it is not proper grammar:
```
WhatNewEpisodes is there any new episodes of {Show}
```

This is an incomplete sentance:
```
WhatNewEpisodes any new episodes of {Show}
```

digiltd:
It is proper grammar "ask kodi is there any new episodes of the simpsons" "is there any new beers to try this week" "is there any apples in the bowl" "is there any old ladies in the pool" 

it's fine :)

and yes it is an incomplete sentence, but that is what i was using to get the information i wanted, i use it a lot and 9 syllables was enough of a mouthful :) and it shouldn't of started playing a random episode either way.

242 was to do with the order of the {show} {season} {episode}

"play season 4 episode 7 of South Park" worked

"play South Park season 4 episode 7" didn't work because there were no utterances for it.

jingai:
http://www.grammar.cl/Present/ThereIsThereAre.htm

I've already explained what was happening with #242.  I'm very aware of how this works and why it fell through the way it did.  What you added is fine, _if_ it doesn't break anything else.  As I noted, I remember playing with alternates, but had some issues with them.  It should be tested is all.

The incomplete sentence is also fine.  Was only noting it because all of the other phrases are designed to make complete sentences that integrate the invocation phrase; e.g., `Alexa, ask Kodi to go get me a beer` rather than something like `Alexa, ask kodi beer`.

digiltd:
yay, arguing about grammar on the internet 

"is there any" "are there any" is very different from "there is" "there are" 

https://english.stackexchange.com/questions/34353/is-there-versus-are-there?noredirect=1&lq=1

https://english.stackexchange.com/questions/17544/is-there-any-proof-versus-are-there-any-proofs?noredirect=1&lq=1

https://www.englishforums.com/English/IsThereAnyVsAreThereAny/vdpvg/post.htm#353324

>e.g., Alexa, ask Kodi to go get me a beer rather than something like Alexa, ask kodi beer.

...but there are plenty of examples in there that are equivalent of "alexa, ask kodi get beer" 

I am not trying to merge a request that changes the whole Skill, just trying to add a few utterances that myself and others would benefit from :)

jingai:
Yes, grammar is kind of important in a program specifically designed to process it.  The links you gave even say you're wrong about it.. but whatever, I guess.

As for utterances in there that don't expand to complete, natural sentences, please point them out and I'll fix them.  Any that are there are not intentional.  Off the top of my head there are indeed a few that I wish didn't exist, such as `big jump backward`, since there's really no natural way to say that.

If you're referring to the lack of `to` in front of everything, it's because I found that Alexa didn't need it.  Contrast this with, `Alexa, ask Kodi if there are any new episodes`, where I did find it needed the `if` in front.

digiltd:
Indeed, "whatever". The links had many examples of both "is there" and "are there" being used, with the majority of responses saying it was correct/acceptable, i'm not going to point each one out.

Yes there were also some (though fewer) that did say it wasn't "correct", but would add that is what people say.

The links certainly don't say "i'm wrong" about it. 

I don't want to nit pick, and wouldn't because of the way utterances work, they don't need to be 100% accurate word for word sentances.. but if you are putting me on the spot, "ask kodi, to suggest show" and "ask kodi to fullscreen" are hardly fully formed sentences :)

Personally imho I have always found "ask kodi to watch..." "ask kodi to listen.." to be really bizarre and something that I wouldn't ever find myself saying. The idea you are asking her to watch something doesn't make sense to me. I know why you use them, you don't need to explain. But breaking it down, it is not "correct", you are not asking Kodi to watch anything, Kodi doesn't have eyes :)

jingai:
IMHO, these are unnecessary (handled by what's there already or just wrong):

```
WhatNewEpisodes is there a new episode of {Show}
WhatNewEpisodes are there any new episodes of {Show}
WhatNewEpisodes are there new episodes of {Show}
WhatNewEpisodes is there new episodes of {Show}
WhatNewEpisodes is there any new episodes of {Show}
```

The rest are fine, if they don't break other things.  If someone else wants to give it a thorough testing, go for it.  If not, I'll do it when I get a chance.

Things like `fullscreen` work with `Alexa, ask Kodi for fullscreen` and thus are fine.  `to` and `for` were liberally omitted for this reason.  `if` for whatever reason wasn't -- at least at one point -- working without adding it explicitly.

`watch` and `listen to` were added as ways to narrow down the media type.  Even Amazon does this.

digiltd:
Not wanting to go on, but if someone tells me I am wrong, I try to understand why, especially when they are so adamant they are correct (and in this instance, I can't let it go). So I looked at your http://www.grammar.cl/Present/ThereIsThereAre.htm link again this morning, on there it gives the example _"Is there any ice-cream in the freezer? - Yes, there is"_ as being a correct sentence. 

Maybe it is a UK/US english difference, in the same way we put the day before the month when giving a date and use a 24hr clock. 

And regarding your points on how _"any new episodes of {Show}"_ is an incomplete sentence and shouldn't be used.  The argument you make about `fullscreen` above, and how you omitted `to` and `for` to allow for the built in "action" words is the same thing, isn't it?

_"ask Kodi for any new episodes of The Flash"_
_"ask Kodi about any new episodes of The Flash"_

Though even in the rawest form _"ask Kodi, any new episodes of The Flash?"_ (note the slight pause) it is a perfectly acceptable and common informal way of asking the same question (in the UK at least)

jingai:
Look, I said "whatever" because I just didn't want to argue it any more.  If you really do care to understand, carefully consider the difference between these:
```
Is there any proof?

Is there any water in the pool?

Are there any new books?
```
All of which are correct; however, these are not:
```
Are there any proof?

Are there any water in the pool?

Is there any new books?
```

Read your own links carefully and you'll see why.  If it's still not obvious, there is one letter that makes the difference.

As for "ask Kodi for any new episodes", there are two things: 1. it might work as-is, and 2. if I asked you for a book, I'd expect you to hand me a book, not tell me that you have one book. 

There may be an argument for adding informal/slang utterances. But I think it's a separate issue to discuss.

jingai:
To push this along a little, I tested all of the show-related Intents after I added [this](https://github.com/jingai/kodi-alexa/tree/watch-ep-variations) to my own model.  Everything is working fine with those -- no collisions or anything.

There _is_ a problem with `WhatNewEpisodes`, but it's not the skill's fault per se.  My previous build of the model handled this utterance just fine:
```
Alexa, ask Kodi if there're any new episodes?
```

The newly-built model has more difficulty expanding that contraction into "there are,' and it instead becomes `if there any new episodes?`, which is something I'd feel dirty adding to the sample utterances lol.. besides, it's probably just a transient issue anyway.

digiltd:
And look, I am not arguing for the sake of being difficult, I have read those and many other pages trying to find a definitive answer. I don't want to be wrong, and happy to accept when I, but the general view is either are acceptable, and I have only ever said "acceptable" or "valid" because there is no simple answer.  

"Are there any new books?" yes

"is there any new books?" not really

"is there any new harry potter books?" yes

"is there any new books in the something series?" yes

though more suited to books

"is there any new chapters of weekly-series?"

I see your point, but in this case, when asking if there is either a single or plural of something that know may or may not exist. It is acceptable.

We are not talking slang, as a google search for the phrase "is there any new" show... and if are talking slang then "give me a hard copy right there" is pretty niche 

I really don't know why this is becoming such an issue, it is only adding is a few utterances based a time when I asked "any new episodes of the flash" (because I had been testing and I got tongue tide, or just plain tired of saying "if there are any new episodes") and it didn't work, it played a random episode.

I am not trying to add "ask kodi if any new Flash dropped this week foshizzle" :)

digiltd:
>Alexa, ask Kodi if there're any new episodes?

didn't work for me

>Alexa, ask Kodi if there're any new episodes of The Flash?

did

Though she heard

image

so i tried again with my best attempt to say "there're'

image

jingai:
Aye, that's what I was stating.  She is currently having trouble with the contraction "there're," but previous builds of the model did not.  Again, it's probably a transient issue, and it's arguable whether or not the fix belongs in the skill.  I really don't want to add "if there any new episodes" to the utterances if we can avoid it.

At this point I don't know how to convince you that "is there any new episodes" is incorrect grammar, so I'm just going to drop it and end with my own opinion: I don't think this belongs in the utterances.  Besides, you can ask this way if you really want to with the current utterances.  It doesn't need to be explicitly added.

As far as adding incomplete sentences, informal language, etc, it should be discussed, but probably some other place than here.

The more informal references that are there for the rather well-known sci-fi scene involving a voice interface to a computer are there for the geek factor.  If someone is taking issue with them, they can be removed -- I doubt anyone _actually_ uses them more than once.  It was added honestly because I wanted to geek out and act out the scene from the movie.. and I figured someone else might want to do the same ;)

digiltd:
I don't think you can convince me that it is categorically wrong and shouldn't ever be used, sorry. 

Well maybe you can, though wrong or not, it is a phrase people use. But as you say, the Skill does work when asked "is there any new..." using the utterances already in place.

This has all been a bit of a distraction from the initial reason for the PR, how "any new episodes of {Show}" failed and played a random episode.

Yes, it is a shortened sentence, but we are talking about giving a command to machine, not speaking to a grandparent :)

I can see you have already created your own branch with the other changes (WatchEpisode, WatchRandomEpisode) so it is not worth me changing mine.

Up to you if you want to add `WhatNewEpisodes any new episodes of {Show}` but currently, if someone does, it doesn't work.

PR #258, opened 2018-03-24
Added delaying/advancing subtitles with voice

aktur:
Now you can delay/advance through Alexa. Default delay is 500ms ("delay/advance subtitles") or you can say "delay/advance subtitles by two seconds". Works better for me than using +/- on remotes as it is not consistent which one delays and which one advances subtitles. With voice its easier and consistent.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,7 +81,9 @@ MUSICVIDEOGENRES
 ADDONS
 
 kodi-alexa/
-run.sh
 
 zappa_settings.json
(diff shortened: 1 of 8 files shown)


m0ngr31:
Can @ausweider take a look at translating this?

danielmasur:
Waiting for your Flask-Ask pip10 compatibility patch to be merged. Don't know how to re-setup my device environment without it/manually applying that fix myself.

Issue #275, opened 2018-12-07
Files i need to translate for Spanish


elopezmadrid:
which files i need to edit?

avogabo:
+1. Help please

jingai:
sample_slotvals.es.txt
templates.es.yaml
utterances.es.txt (and convert to speech_assets/SampleUtterances.es.txt)

ruben0909:
@jingai Could you review this PR 
https://github.com/m0ngr31/kanzi/pull/276
and this
https://github.com/m0ngr31/lexigram-cli/pull/4
Thank you in advance

PR #276, opened 2019-04-14
Add Spanish

ruben0909:
Templates, Samples, utterances
Full translaction to spanish
Without accents and without ñ because of http://www.makermusings.com/amazon-echo-utterance-expander/ doesn´t allow it.


ruben0909:
This resolve #275

m0ngr31:
Looks good :)

m0ngr31:
> Without accents and without ñ because of http://www.makermusings.com/amazon-echo-utterance-expander/ doesn´t allow it.

Maybe I need to build a new expander that supports that...

m0ngr31 (approved):

ruben0909:
Thank you

PR #286, opened 2020-03-16
fix wrong assignments from ViewEpisodePlaylistRecent to ViewMusicVide…

skaaj4:
ViewEpisodePlaylistRecent intent uses utterances for ViewMusicVideoPlaylistRecent in the german language. Because the custom intent ViewMusicVideoPlaylistRecent has no utterances at all, the deploying thows warnings.


jingai (approved):
Looks good to me.

PR #291, opened 2022-03-18
Fix 'requirements.txt' git:// repo URIs (unsupported), switch to https://.

pataquets:
While working on making deploy to Heroku work again (PR to follow soon), build fails due to `git://` protocol no longer supported. Just switching requirements from `git://` to `https://` and worked like a charm.
@m0ngr31 : After testing it a bit, I'll send a fix to Heroku deployment, which is currently failing. Stay tuned :grin:.


pataquets:
@m0ngr31 : also submitted #292, which requires this PR previously.

PR #294, opened 2022-05-17
Fix the 'Deploy to Heroku' button.

pataquets:
Although the deploy to Heroku succeeds, you need to manually set the stack for the application:
``` shell
heroku stack:set heroku-18
```
Since the `heroku-18` stack is now required, set it explicitly in the `app.json` file for the 'Deploy to Heroku' button to work.

### DIFF
--- a/app.json
+++ b/app.json
@@ -123,5 +123,6 @@
       "required": false
     }
   },
-  "image": "heroku/python"
+  "image": "heroku/python",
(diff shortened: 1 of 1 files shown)


pataquets:
@m0ngr31 button is still failing, even when manual deployment already works.

pataquets:
@m0ngr31 ping

